PR #3, opened 2017-02-03
added sample usable as-is from jenkins image; add create from http; f…

gabemontero:
…ix selector related bug

@jupierce @bparees fyi / ptal

I plan on pulling the the jenkins-image-sample.groovy into our jenkins image as a second sample job to accompany the current sample job we have to leverages our traditional plugin example.

### DIFF
--- a/examples/jenkins-image-sample.groovy
+++ b/examples/jenkins-image-sample.groovy
@@ -1,0 +1,121 @@
+/**
+ * This script does nothing in particular,
+ * but is meant to show actual usage of most of the API.
+ */
+
+openshift.withCluster() {
+
+    def saSelector1 = openshift.selector( "serviceaccount" )
+    saSelector1.describe()
+    
+    timeout(3) {
+        def templateSelector = openshift.selector( "template", "mongodb-ephemeral")
+        
+        templateSelector.delete('--ignore-not-found')
+        // secrets are not covered by the oc all qualifier, so have to handle separately
+        openshift.delete('all,secret', '-l', 'template=mongodb-ephemeral-template', '--ignore-not-found')
+        template = openshift.create('https://raw.githubusercontent.com/openshift/origin/master/examples/db-templates/mongodb-ephemeral-template.json').object()
+        
+        // Explore the Groovy object which models the OpenShift template as a Map
+        echo "Template contains ${template.parameters.size()} parameters"
+    
+        // For fun, modify the template easily while modeled in Groovy
+        template.labels["mylabel"] = "myvalue"
+    
+        // Process the modeled template. We could also pass JSON/YAML or a template name instead.
+        // note: -p option for oc process not in the oc version that we currently ship with openshift jenkins images
+        def objectModels = openshift.process( template )//, "-p", "MEMORY_LIMIT=600Mi")
+    
+        // objectModels is a list of objects the template defined, modeled as Groovy objects
+        echo "The template will create ${objectModels.size()} objects"
+    
+        // For fun, modify the objects that have been defined by processing the template
+        for ( o in objectModels ) {
+            o.metadata.labels[ "anotherlabel" ] = "anothervalue"
+        }
+    
+        // Serialize the objects and pass them to the create API.
+        // We could also pass JSON/YAML directly; openshift.create(readFile('some.json'))
+        def created = openshift.create(objectModels)
+    
+        // Create returns a selector which will always select the objects created
+        created.withEach {
+            // Each loop binds the variable 'it' to a selector which selects a single object
+            echo "Created ${it.name()} from template with labels ${it.object().metadata.labels}"
+        }
+    
+        // Filter created objects and create a selector which selects only the new DeploymentConfigs
+        def dcs = created.narrow("dc")
+        echo "Database will run in deployment config: ${dcs.name()}"
+        timeout(5) {
+            // Find a least one pod related to the DeploymentConfig and wait it satisfies a condition
+            dcs.related('pods').untilEach(1) {
+                // untilEach only terminates when each selected item causes the body to return true
+                return it.object().status.phase != 'Pending'
+            }
+        }
+    
+        // Print out all pods created by the DC
+        echo "Template created pods: ${dcs.related('pods').names()}"
+    
+        // Show how we can use labels to select as well
+        echo "Finding dc using labels instead: ${openshift.selector('dc',[mylabel:'myvalue']).names()}"
+    
+        echo "DeploymentConfig description"
+        dcs.describe()
+        echo "DeploymentConfig history"
+        dcs.rollout().history()
+    
+        def rubySelector = openshift.selector("bc", "ruby")
+        def builds
+        try {
+            rubySelector.object()
+            builds = rubySelector.related( "builds" )
+        } catch (Throwable t) {
+            // The selector returned from newBuild will select all objects created by the operation
+            nb = openshift.newBuild( "https://github.com/openshift/ruby-hello-world", "--name=ruby" )
+    
+            // Print out information about the objects created by newBuild
+            echo "newBuild created: ${nb.count()} objects : ${nb.names()}"
+            
+            // Filter non-BuildConfig objects and create selector which will find builds related to the BuildConfig
+            builds = nb.narrow("bc").related( "builds" )
+    
+        }
+    
+        // Raw watch which only terminates when the closure body returns true
+        builds.watch {
+            // 'it' is bound to the builds selector.
+            // Continue to watch until at least one build is detected
+            if ( it.count() == 0 ) {
+                return false
+            }
+            // Print out the build's name and terminate the watch
+            echo "Detected new builds created by buildconfig: ${it.names()}"
+            return true
+        }
+    
+        echo "Waiting for builds to complete..."
+    
+        // Like a watch, but only terminate when at least one selected object meets condition
+        builds.untilEach {
+            return it.object().status.phase == "Complete"
+        }
+    
+        // Print a list of the builds which have been created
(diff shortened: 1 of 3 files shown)


bparees:
I don't think we want to actually add it as a sample job in the jenkins image. I kinda regret putting in the sample job that's already there as it is.  I think we're better served by external tutorials that make people create the job themselves since the sample job is just a distraction for anyone who's not a first time deployer of jenkins, so adding a second one is, I think, the wrong direction.  

But having it as a usable documented example definitely makes sense.

gabemontero:
10-4 I'll incorporate it with our jenkins pipeline samples under
origin/example/jenkins instead

gabemontero:
Moving off of how to expose the out of the box example, @jupierce - are you OK with the code changes that stemmed from crafting the example?  You OK if I hit the merge button?

examples/jenkins-image-sample.groovy
```diff
@@ -0,0 +1,119 @@
...
+    def saSelector1 = openshift.selector( "serviceaccount" )
+    saSelector1.describe()
+    def saSelector2 = openshift.selector( "sa/builder" )
+    def saSelector3 = openshift.selector( "sa", "builder" )
+    def saSelector4 = openshift.selector( "dc", [ "template" : "mytemplate" ] )
+    
+    def templateSelector = openshift.selector( "template", "mongodb-ephemeral")
+        
+    templateSelector.delete('--ignore-not-found')
+    openshift.delete('all,secret', '-l', 'template=mongodb-ephemeral-template', '--ignore-not-found')
```
jupierce: This looks like a good point to use the "related" API.

examples/jenkins-image-sample.groovy
```diff
@@ -0,0 +1,119 @@
...
+
+    echo "DeploymentConfig description"
+    dcs.describe()
+    echo "DeploymentConfig history"
+    dcs.rollout().history()
+
+    def rubySelector = openshift.selector("bc", "ruby")
+    def builds
+    try {
+        rubySelector.object()
```
jupierce: Is it possible to handle this without catching exceptions? Since this is educational, I'd like it to propose a best practice.
```
def builds = rubySelector.related('builds') // This can be created even if bc/ruby doesn't exist yet
if ( rubySelector.count() == 0 ) { 
...newBuild...
}
```

You might even consider eliminating reuse altogether in this sample by deleting the bc/builds. The comment at the top says the test cleans up after itself.

examples/jenkins-image-sample.groovy
```diff
@@ -0,0 +1,119 @@
...
+        // Print out information about the objects created by newBuild
+        echo "newBuild created: ${nb.count()} objects : ${nb.names()}"
+        
+        // Filter non-BuildConfig objects and create selector which will find builds related to the BuildConfig
+        builds = nb.narrow("bc").related( "builds" )
+
+    }
+
+    // Raw watch which only terminates when the closure body returns true
+    builds.watch {
```
jupierce: Even though I fail to in my original coverage.groovy, I recommend any watch/untilEach be wrapped wtihin a timeout closure. You can have one large one at the top that encompasses the whole pipeline if you want.

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -463,10 +463,11 @@ class OpenShiftDSL implements Serializable {
         boolean markup = false;
         String s = obj.toString();
         markup = s.contains("{") || s.contains(":"); // does this look like json or yaml?
+        boolean httpref = s.startsWith("http") && verb.equalsIgnoreCase("create"); // a create from a http or https raw.githubuser path
```
jupierce: toLowercase before startsWith?

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -709,7 +713,9 @@ class OpenShiftDSL implements Serializable {
                 kind = s[0];
                 qualifier = s[1];
             }
-
+            
+            if (abbreviations.containsKey(kind))
```
jupierce: Recommending braces on this if statement.

jupierce:
@gabemontero Submitted a review for your consideration, but feel free to merge.

examples/jenkins-image-sample.groovy
gabemontero: Sorry, based on what I'm seeing in the `related()` code and the other `related()` examples, I'm missing the exact change you think would be good here.

Can you clarify please?

examples/jenkins-image-sample.groovy
gabemontero: I originally tried the count == 0 approach but the code seemed to hang.  I'll try that again first and report on the findings.

Then I'll circle back to deleting bc/builds.

gabemontero: Yeah i doesn't work, but I misinterpreted the hang.

Even though there are no bc's yet in my project (confirmed via use of `oc` from where I ran `oc cluster up`, `oc new-app jenkins-ephemeral`, etc.), `rubySelector.count()` is returning 1.  When I call names to dump the output, it returns `buildconfig/ruby` .

The *hang* stems from the subsequent build watch, as `new-build` was not called and there in fact is not a *ruby* bc yet.  It occurs even if I move the related call after the count call.

I could add debug to confirm, but at a glance, I suspect the `objectList` for the `rubySelector` is being populated on the constructor here:  https://github.com/openshift/jenkins-client-plugin/blob/master/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy#L718

So should I just check for less than or equal to 1 on the count, or should a fix be made to ensure count == 0 when a selector is created ?

examples/jenkins-image-sample.groovy
gabemontero: sure ... any thoughts on what the timeout should be? .... somewhere around 3 to 5 minutes for the whole job came to mind.

gabemontero: well ... just noticed the timeout of 5 for the db deploy ... going to go with 7 for the overall job, unless you think different

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
gabemontero: part of next push

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
gabemontero: part of next push

examples/jenkins-image-sample.groovy
jupierce: Instead of
```openshift.delete('all,secret', '-l', 'template=mongodb-ephemeral-template', '--ignore-not-found')```
something like:
```
templateSelector.related('all').delete('--ignored-not-found')
templateSelector.related('secret').delete('--ignored-not-found')
```

Comment (gabemontero) [inline examples/jenkins-image-sample.groovy]: ah -- cool thanks ... will update

File: examples/jenkins-image-sample.groovy (hunk 2 above)
Comment (jupierce) [inline examples/jenkins-image-sample.groovy]: d'oh. You're right. From OpenShiftGlobalVariable/help.jelly :

> Returns the number of objects the receiver selects. In the case of a DynamicSelector,
> a query is made to the server to establish the current count. **In this case of StaticSelectors, this count is a constant (even if the objects it selects do not exist on the server).**

In other words. selector('bc','ruby').count() will always be 1. So... perhaps a API for this use case?  
```selector('bc','ruby').exists()```
And/or an optional flag on count() to force a static selector to do a query to the server?
```selector('bc','ruby').count(true)```

Comment (jupierce) [inline examples/jenkins-image-sample.groovy]: I was just playing around with trying to implement exists() or count(onlyLiveObjects:true) . The problem is that any get operation that specifies an exact name:
```
oc get bc/ruby
```
will always return an error if the object doesn't exist. I don't see an --ignore-not-found type flag for oc get. Unless we scan the stderr text for the nature of the error, we won't know whether the object didn't exist or, for example, there was an error communicating with the server. 

Maybe someone has a trick for this?

examples/jenkins-image-sample.groovy
gabemontero: Unfortunately it does not work.  This does not seem to anything over or above what the templateSelector.delete did.

Switching back to what I had for now.  If time permits, I'll circle back to this and see if I can make plugin changes to have this work.

examples/jenkins-image-sample.groovy
gabemontero: Not using the `oc` command, no.  Pending an update to `oc` to better accommodate this, and assuming we don't want to scan stderr for arbitrary text, does this use case warrant making the HTTP rest call and interpreting the  HTTP rc's in order to implement?

gabemontero:
I've pushed a portion of @jupierce 's suggestions.  What was not included:
- using `templateSelector.related(...).delete(...)` instead of `openshift.delete(...)`  ... use of related did not do what was hoped; I'll do a bit more debug before calling it with this PR

- the selector exists notion which @jupierce already noted the challenge with.... pending any decisions around REST calls or parsing stderr, leaving sample as is.  Note, I removed the cleanup portion of the comment up top, and want to leave the current try / catch snippet in vs. cleaning up at the end.  Even if it is not stylistic from a code appearance perspective, I think it is important to note what is required for *existence* checking. 

Will merge on full completion of those 2 bullets per @jupierce 's early approval.  Thanks @jupierce .

PR #20, opened 2017-04-28
add test for list to str array conversion for params

gabemontero:
@jupierce @jcpowermac FYI ... the PR test for the recent fix from @jupierce


jupierce:
I might recommend all three supported variations:
1. The one you already have
2. One where a real Object[] or String[] is passed in
3. One where a standard list of arguments is passed in

gabemontero:
test case updated @jupierce

gabemontero:
tests passed ... merging

note, I did confirm locally that the tests failed before @jupierce 's fix from yesterday

PR #21, opened 2017-05-11
add selector exists; tweak to timeout, try/catch error handling

gabemontero:
@bparees @jupierce FYI

- eventually, when we pull in a 3.6 `oc` binary into the openshift jenkins image, I'll start the ball rolling the approved version of this pull

- given the current workflow, I'll un-comment the test case change after cutting a new version of the client plugin - don't want to make the new test case available to the overnight tests before the new plugin version is available in the jenkins image (longer term, need to see how we can only expose new test cases once their associated plugin versions are available

Otherwise, comment as desire.  Disclaimer - I don't expect to have say both forms of `exist` survive long term.  I left these various options in to foster discussion.

### DIFF
--- a/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
+++ b/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
@@ -4,14 +4,20 @@ import com.cloudbees.groovy.cps.NonCPS
 import com.cloudbees.plugins.credentials.CredentialsProvider
 import com.openshift.jenkins.plugins.pipeline.OcContextInit
 import com.openshift.jenkins.plugins.pipeline.OcAction
+
 import groovy.json.JsonOutput
 import groovy.json.JsonSlurper
 import hudson.AbortException
 import hudson.FilePath
 import hudson.Util
 
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
 
 class OpenShiftDSL implements Serializable {
+    
+    static final Logger LOGGER = Logger.getLogger(OpenShiftDSL.class.getName());
 
     private org.jenkinsci.plugins.workflow.cps.CpsScript script
 
@@ -19,11 +25,35 @@ class OpenShiftDSL implements Serializable {
     private transient OpenShift.DescriptorImpl config = new OpenShift.DescriptorImpl();
 
     private int logLevel = 0; // Modified by calls to openshift.logLevel
+    
+    private HashMap<String,Capabilities> nodeCapabilities = new HashMap<String,Capabilities>();
 
     public OpenShiftDSL(org.jenkinsci.plugins.workflow.cps.CpsScript script) {
         this.script = script
     }
 
+    public synchronized Capabilities getCapabilities() {
+          String key = script.env.NODE_NAME
+          Capabilities caps = nodeCapabilities.get(key)
+          if ( caps != null ) {
+                return caps
+          } else {
+                caps = new Capabilities()
+                ArrayList<String> g = new ArrayList<String>();
+                g.add("get");
+                OcAction.OcActionResult versionCheck = (OcAction.OcActionResult)script._OcAction( buildCommonArgs("help", g, null, null) );
+                LOGGER.log(Level.FINE, "getCapabilities return from oc help get " + versionCheck.out);
+                if (versionCheck.out.contains("--ignore-not-found")) {
+                    caps.ignoreNotFound = true;
+                } else {
+                    caps.ignoreNotFound = false;
+                }
+                nodeCapabilities.put( key, caps )
+                LOGGER.log(Level.FINE, "getCapabilities nodeCapabilites: " + nodeCapabilities);
+                return caps
+          }
+    }
+    
     private Context currentContext = null;
 
     private static final Map<String,String> abbreviations = [
@@ -759,6 +789,18 @@ class OpenShiftDSL implements Serializable {
         }
 
     }
+    
+    public class Capabilities implements Serializable {
+        private boolean ignoreNotFound;
+        
+        public Capabilities() {
+        }
+
+        public boolean hasIgnoredNotFound() {
+            return ignoreNotFound;
+        }
+        
+    }
 
     public class OpenShiftResourceSelector extends Result implements Serializable {
 
@@ -900,6 +942,14 @@ class OpenShiftDSL implements Serializable {
             }
         }
 
+        public boolean exists() throws AbortException {
+            if ( objectList != null ) {
+                  // If object names are explicitly given, make sure they *all* exists
+                  return count() == objectList.size()
+            }
+            return count() > 0 
+        }
+
         public void untilEach( int min=1, Closure<?> body ) {
             watch {
                 while ( true ) {
@@ -935,7 +985,7 @@ class OpenShiftDSL implements Serializable {
         }
 
         public int count() throws AbortException {
-            return names().size();
+            return queryNames().size();
         }
 
         public HashMap object(Map mode=null) throws AbortException {
@@ -951,17 +1001,31 @@ class OpenShiftDSL implements Serializable {
             }
             return m;
         }
+        
+        private ArrayList<String> queryNames() throws AbortException {
+            // Otherwise, we need to ask the API server what presently matches
+            OcAction.OcActionResult r = null;
+            if (script.openshift.getCapabilities().hasIgnoredNotFound()) {
+                r = (OcAction.OcActionResult)script._OcAction( buildCommonArgs("get", selectionArgs(), null, "-o=name", "--ignore-not-found") );
+                r.failIf( "Unable to retrieve object names: " + this.toString() );
+            } else {
+                r = (OcAction.OcActionResult)script._OcAction( buildCommonArgs("get", selectionArgs(), null, "-o=name") );
+                if ( r.status != 0 && r.err.contains("(NotFound)") ) {
+                    return new ArrayList<String>();
+                } else {
+                    r.failIf( "Unable to retrieve object names: " + this.toString() );
+                }
+            }
+
+            return OpenShiftDSL.splitNames( r.out );
+
+        }
 
         public ArrayList<String> names() throws AbortException {
             if ( objectList != null ) {
                 return objectList;
             }
-
-            // Otherwise, we need to ask the API server what presently matches
-            OcAction.OcActionResult r = (OcAction.OcActionResult)script._OcAction( buildCommonArgs("get", selectionArgs(), null, "-o=name") );
-            r.failIf( "Unable to retrieve object names: " + this.toString() );
-
-            return OpenShiftDSL.splitNames( r.out );
+            return queryNames();
         }
 
         public String name() throws AbortException {
@@ -1024,8 +1088,7 @@ class OpenShiftDSL implements Serializable {
         public Result cancelBuild(Object... userArgs ) throws AbortException {
             return onceForEach( "cancelBuild", "cancel-build", userArgs );
         }
-
-
+        
         public Result deploy(Object... userArgs ) throws AbortException {
             return onceForEach( "deploy", "deploy", userArgs );
         }
--- a/src/main/resources/com/openshift/jenkins/plugins/pipeline/OpenShiftGlobalVariable/help.jelly
+++ b/src/main/resources/com/openshift/jenkins/plugins/pipeline/OpenShiftGlobalVariable/help.jelly
@@ -740,6 +740,25 @@
             </dd>
 
             <dt>
+                <a name="Selector_exists"/>
+                <code>Selector.exists():boolean</code><br></br>
+            </dt>
+            <dd>
+                <p>
+                    <p style="margin-left: 1em; color:#657383;">Example:
+                        <code>
+                            // Returns a boolean whether the template "mongodb-ephemeral is present in the project.<br></br>
+                            def n = openshift.selector( "template", "mongodb-ephemeral" ).exists() <br></br>
+                        </code>
+                    </p>
+                    <br></br>
+                    Returns true if a non-zero number of objects based on the selection criteria exits. In the case of a DynamicSelector,
+                    a query is made to the server to establish existence. In the case of StaticSelectors,
+                    the entire list of objects supplied in the creation of the selector must be present for a return value of "true" to occur.
+                </p>
+            </dd>
+
+            <dt>
                 <a name="Selector_count"/>
                 <code>Selector.count():Integer</code><br></br>
             </dt>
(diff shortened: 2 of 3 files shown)


jupierce:
My first thought is that the closure form doesn't offer much over a simple `boolean Selector.exists()`


```
 public boolean exists() throws AbortException {
            return names().size() > 0
}
```

This would be possible if --ignore-not-found is always specified in the call to `oc get` from names().  Putting it in names() would simultaneously fix count() and other functions when no results were returned.


I would expect this to be the usage common pattern:
```
while ( ! sel.exists() ) {
    sleep 60
}
```

jupierce:
To ensure backwards compatibility, it would also be a great time to introduce a version check into the codebase.

names() would use --ignore-not-found when the plugin detects it is running a version which supports it. When not running a supported version, --ignore-not-found would not be used and stderr would be inspected for the not found string.

gabemontero:
thanks @jupierce - first blush I'm convinced wrt each of your points.  And as an aside, I feel validated in that my sense of uncertainty on which method signature to employ was spot on :-)

It will probably be a little bit before I circle back to this given the new sprint starting and some overall 3.6 jenkins to-do's still pending.

gabemontero:
move of relevant logic to `names()`, remove of closures, addition of version check, associated test change,  pushed

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -900,6 +910,17 @@ class OpenShiftDSL implements Serializable {
             }
         }
 
+        public boolean exists() throws AbortException {
+            return names().size() > 0
+        }
+
+        public void notExists( Closure<?> body ) {
```
jupierce: Did you mean to keep this method around?  I don't think the implementation will work.

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -900,6 +910,17 @@ class OpenShiftDSL implements Serializable {
             }
         }
 
+        public boolean exists() throws AbortException {
+            return names().size() > 0
```
jupierce: I was wrong here since this implementation will not work with the following:

```
s = openshift.selector( 'pod/x', 'pod/x' )
s.exists()   // This will always return true with the current PR because it creates a static objectList based selector.
```

Instead, we would need a new method:
```
// New method which always queries for names even if the selector is a static list
private queryNames() throws AbortException {
      ...content of current names(), but no opening      if ( objectList != null )   check
}

public ArrayList<String> names() throws AbortException {
        if ( objectList != null ) {
                 return objectList;
        }
        return queryNames()
}

// Tweak to count's implementation
public int count()  throws AbortException {
        return queryNames().size()
}

public boolean exists()  throws AbortException {
         if ( objectList != null ) {
               // If object names are explicitly given, make sure they *all* exists
               return count() == objectList.length
         }
         return count() > 0 
}

``

File: src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy (hunk 1 above)
Comment (gabemontero) [inline src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy]: Nope forgot to remove.  Thanks.

File: src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy [hunk 3]
```diff
@@ -956,10 +977,20 @@ class OpenShiftDSL implements Serializable {
...
             }
-
+            
             // Otherwise, we need to ask the API server what presently matches
-            OcAction.OcActionResult r = (OcAction.OcActionResult)script._OcAction( buildCommonArgs("get", selectionArgs(), null, "-o=name") );
-            r.failIf( "Unable to retrieve object names: " + this.toString() );
+            OcAction.OcActionResult r = null;
+            ArrayList<String> g = new ArrayList<String>();
+            g.add("get")
+            OcAction.OcActionResult versionCheck = (OcAction.OcActionResult)script._OcAction( buildCommonArgs("help", g, null, null) );
```
gabemontero: Yeah I entertained that notion as well.  Ultimately though, I think we need to do this periodically instead of just onece since the user could import different versions of `oc` through Global Tool Configuration.

I then decided to keep it simple for now.  I should minimally have put a TOD comment referring to such an optimization.  Unless we have further debate here, I'll do that with the next push.

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -956,10 +977,20 @@ class OpenShiftDSL implements Serializable {
...
+            LOGGER.log(Level.FINE, "oc help get output: " + versionCheck.out)
+            if (versionCheck.out.contains("--ignore-not-found")) {
+                r = (OcAction.OcActionResult)script._OcAction( buildCommonArgs("get", selectionArgs(), null, "-o=name", "--ignore-not-found") );
+                r.failIf( "Unable to retrieve object names: " + this.toString() );
+            } else {
+                r = (OcAction.OcActionResult)script._OcAction( buildCommonArgs("get", selectionArgs(), null, "-o=name") );
+                r.failIf( "Unable to retrieve object names: " + this.toString() );
```
gabemontero: ok thanks

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
gabemontero: I hadn't previously picked up on this "static objectList based selector" notion from the code or associated comments, nor am I sure I'm fully gathering it from you explanation here.  Could you elaborate (either in this comment thread or face to face if you have a sec) ?

That said, the null checks for `objectList` and this snippet from the selector constructor did catch my eye previously:

```
            if ( qualifier != null ) {
                if ( qualifier instanceof Map ) {
                    this.labels = new HashMap((Map)qualifier);
                } else {
                    // Otherwise, the qualifier is a name that is paired with the name
                    objectList = Arrays.asList( kind + "/" + qualifier.toString() );
                    kind = null;
                }
            }

```

and was curious if that affected the previous suggested solution.

I'm thinking that is somewhat involved here then with what you found.  But I'd still like to get a better grasp of your terminology, etc.

All that said, I'll push up a new version with these changes later today.

Thanks.

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
jupierce: The suggestion I included should re-acquire the capabilities of a client every time a pipeline runs (and so should cover the case of the tool configuration changes from run to run). It will also acquire it for every different node (in case you have heterogeneous client installs).

gabemontero: duh ... yeah, it is an instance var, and a new selector is constructed each time; was distracted by something else, missed the obvious

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
gabemontero: though `g` in the current form contains 1 entry .... we don't want to do that 

i'll return an empty array list

gabemontero:
ok i've pushed the update based on @jupierce 's last set of comments / assistance

also, per our (both @bparees and myself) recent experiences with the ci.openshift pipeline extended test runs, I've incorporated some general best practice suggestions re: try/catch (from online doc and my notes from jenkins world) along with bump/consolidation of time out (based on ci.openshift results and general time frames for extended tests).

jupierce:
Just a reminder to add doc for the new APIs: https://github.com/openshift/jenkins-client-plugin/blob/master/src/main/resources/com/openshift/jenkins/plugins/pipeline/OpenShiftGlobalVariable/help.jelly

gabemontero:
On Tue, May 23, 2017 at 3:53 PM, Justin Pierce <notifications@github.com>
wrote:

> Just a reminder to add doc for the new APIs: https://github.com/openshift/
> jenkins-client-plugin/blob/master/src/main/resources/com/
> openshift/jenkins/plugins/pipeline/OpenShiftGlobalVariable/help.jelly
>
Thanks for catching that.  Update pushed.

src/main/resources/com/openshift/jenkins/plugins/pipeline/OpenShiftGlobalVariable/help.jelly
```diff
@@ -739,6 +739,25 @@
             </dd>
 
             <dt>
+                <a name="Selector_exists"/>
+                <code>Selector.count():boolean</code><br></br>
```
jupierce: .exists()

gabemontero:
Updates for both items pushed

gabemontero:
bump @jupierce  - are we good to go?

gabemontero:
cool ... thanks for all the help @jupierce !!

gabemontero:
delay / hiccup in mongo deploy, slow image pull, test timed out with old timeout values:

```
May 25 10:24:17.953: INFO: At 2017-05-25 10:15:02.356568828 -0400 EDT - event for mongodb-1-deploy: {kubelet 172.18.4.201} Pulled: Container image "openshift/origin-deployer:b8a4a2c" already present on machine
May 25 10:24:17.953: INFO: At 2017-05-25 10:15:04.055410638 -0400 EDT - event for mongodb-1-deploy: {kubelet 172.18.4.201} Created: Created container with id c1b814e8b802353c863adb5d14306c34e91ff986e198ca0b2764980880c37c2b
May 25 10:24:17.953: INFO: At 2017-05-25 10:15:04.625408918 -0400 EDT - event for mongodb-1-deploy: {kubelet 172.18.4.201} Started: Started container with id c1b814e8b802353c863adb5d14306c34e91ff986e198ca0b2764980880c37c2b
May 25 10:24:17.953: INFO: At 2017-05-25 10:15:06.55412747 -0400 EDT - event for mongodb-1: {replication-controller } SuccessfulCreate: Created pod: mongodb-1-97r53
May 25 10:24:17.953: INFO: At 2017-05-25 10:15:06.565675254 -0400 EDT - event for mongodb-1-97r53: {default-scheduler } Scheduled: Successfully assigned mongodb-1-97r53 to 172.18.4.201
May 25 10:24:17.953: INFO: At 2017-05-25 10:16:38.402247936 -0400 EDT - event for mongodb-1-97r53: {kubelet 172.18.4.201} Pulling: pulling image "centos/mongodb-32-centos7@sha256:77b341d1fa2738f03a79a66b857cc363bdad3dec4a6bc60e73ec47c773762cd1"
May 25 10:24:17.953: INFO: At 2017-05-25 10:20:12.596889049 -0400 EDT - event for mongodb-1-97r53: {kubelet 172.18.4.201} Pulled: Successfully pulled image "centos/mongodb-32-centos7@sha256:77b341d1fa2738f03a79a66b857cc363bdad3dec4a6bc60e73ec47c773762cd1"
May 25 10:24:17.953: INFO: At 2017-05-25 10:21:28.499505779 -0400 EDT - event for mongodb-1-97r53: {kubelet 172.18.4.201} Created: Created container with id 88b3c0ef9d1e4707ff000486b8aeab483737a1e52d30fa20645c37b1ab2a335e
May 25 10:24:17.953: INFO: At 2017-05-25 10:22:20.644970223 -0400 EDT - event for mongodb-1-97r53: {kubelet 172.18.4.201} Started: Started container with id 88b3c0ef9d1e4707ff000486b8aeab483737a1e52d30fa20645c37b1ab2a335e
May 25 10:24:17.953: INFO: At 2017-05-25 10:23:26.875846905 -0400 EDT - event for mongodb-1-97r53: {kubelet 172.18.4.201} Killing: Killing container with id docker://88b3c0ef9d1e4707ff000486b8aeab483737a1e52d30fa20645c37b1ab2a335e:pod "mongodb-1-97r53_extended-test-jenkins-pipeline-xzbb0-v6gbx(8d85ad52-4154-11e7-b7de-0e280a171aee)" container "mongodb" is unhealthy, it will be killed and re-created.
May 25 10:24:17.953: INFO: At 2017-05-25 10:23:26.885293638 -0400 EDT - event for mongodb-1-97r53: {kubelet 172.18.4.201} Pulled: Container image "centos/mongodb-32-centos7@sha256:77b341d1fa2738f03a79a66b857cc363bdad3dec4a6bc60e73ec47c773762cd1" already present on machine
May 25 10:24:17.957: INFO: POD                                          NODE          PHASE      GRACE  CONDITIONS
May 25 10:24:17.958: INFO: docker-registry-1-91680                      172.18.4.201  Running           [{Initialized True 0001-01-01 00:00:00 +0000 UTC 2017-05-25 10:11:10 -0400 EDT  } {Ready True 0001-01-01 00:00:00 +0000 UTC 2017-05-25 10:11:20 -0400 EDT  } {PodScheduled True 0001-01-01 00:00:00 +0000 UTC 2017-05-25 10:11:10 -0400 EDT  }]
May 25 10:24:17.958: INFO: router-2-ttfj2                               172.18.4.201  Running           [{Initialized True 0001-01-01 00:00:00 +0000 UTC 2017-05-25 10:11:36 -0400 EDT  } {Ready True 0001-01-01 00:00:00 +0000 UTC 2017-05-25 10:11:56 -0400 EDT  } {PodScheduled True 0001-01-01 00:00:00 +0000 UTC 2017-05-25 10:11:36 -0400 EDT  }]
May 25 10:24:17.958: INFO: jenkins-1-f4jb3                              172.18.4.201  Running           [{Initialized True 0001-01-01 00:00:00 +0000 UTC 2017-05-25 10:13:08 -0400 EDT  } {Ready False 0001-01-01 00:00:00 +0000 UTC 2017-05-25 10:23:51 -0400 EDT ContainersNotReady containers with unready status: [jenkins]} {PodScheduled True 0001-01-01 00:00:00 +0000 UTC 2017-05-25 10:13:08 -0400 EDT  }]
May 25 10:24:17.958: INFO: jenkins-client-plugin-snapshot-test-1-build  172.18.4.201  Succeeded         [{Initialized True 0001-01-01 00:00:00 +0000 UTC 2017-05-25 10:11:38 -0400 EDT PodCompleted } {Ready False 0001-01-01 00:00:00 +0000 UTC 2017-05-25 10:13:02 -0400 EDT PodCompleted } {PodScheduled True 0001-01-01 00:00:00 +0000 UTC 2017-05-25 10:11:38 -0400 EDT  }]
May 25 10:24:17.958: INFO: mongodb-1-97r53                              172.18.4.201  Running           [{Initialized True 0001-01-01 00:00:00 +0000 UTC 2017-05-25 10:15:06 -0400 EDT  } {Ready False 0001-01-01 00:00:00 +0000 UTC 2017-05-25 10:15:06 -0400 EDT ContainersNotReady containers with unready status: [mongodb]} {PodScheduled True 0001-01-01 00:00:00 +0000 UTC 2017-05-25 10:15:06 -0400 EDT  }]
May 25 10:24:17.958: INFO: mongodb-1-deploy                             172.18.4.201  Running           [{Initialized True 0001-01-01 00:00:00 +0000 UTC 2017-05-25 10:15:00 -0400 EDT  } {Ready True 0001-01-01 00:00:00 +0000 UTC 2017-05-25 10:15:04 -0400 EDT  } {PodScheduled True 0001-01-01 00:00:00 +0000 UTC 2017-05-25 10:15:00 -0400 EDT  }]
May 25 10:24:17.958: INFO: 
May 25 10:24:17.960: INFO: 

```

Given prior successful test run, and that last update was to help.jelly, merging the pull

PR #29, opened 2017-05-31
expand use of dtc exit status workaround

gabemontero:
have seen some more intermittent failures in this space in various extended test runs on origin, where the stack traces singled out the places we were still calling dtc.exitStatus without our try/catch wrappering

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/OcAction.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/OcAction.java
@@ -46,20 +46,21 @@ public class OcAction extends AbstractStepImpl {
         this.reference = reference==null?(new HashMap<String,String>()):reference;
     }
 
-    public static void exitStatusRaceConditionBugWorkaround( Controller dtc, FilePath filePath, Launcher launcher ) throws InterruptedException, IOException {
+    public static Integer exitStatusRaceConditionBugWorkaround( Controller dtc, FilePath filePath, Launcher launcher ) throws InterruptedException, IOException {
         for ( int tries = 30; tries > 0; tries-- ) {
             try {
                 // exitStatus can throw an IOException (reporting a NumberFormatException) if the PID file has been created but not
-                // populated. Make sure it stops throwing this exception before continuing.
-                dtc.exitStatus(filePath,launcher);
-                return; // exitStatus is healthy, we should be safe
+                // populated, or if the PID file has not yet been created. Make sure it stops throwing this exception before continuing.
+                return dtc.exitStatus(filePath,launcher);
             } catch ( IOException ioe ) {
                 if ( tries == 1 ) {
                     throw ioe;
                 }
-                Thread.sleep(1000);
+                Thread.sleep(125);
             }
         }
+        launcher.getListener().getLogger().println("After 30 retries, unable to get exit status for " + filePath.toURI().toString());
+        return -1;
     }
 
     public static class OcActionResult implements Serializable {
@@ -195,11 +196,9 @@ public class OcAction extends AbstractStepImpl {
                 ByteArrayOutputStream stdErr = new ByteArrayOutputStream();
                 ByteArrayOutputStream stdOut = new ByteArrayOutputStream();
 
-                exitStatusRaceConditionBugWorkaround( dtc, filePath, launcher);
-
                 long reCheckSleep = 250;
                 Integer exitStatus;
-                while ( ( exitStatus = dtc.exitStatus(filePath,launcher) ) == null ) {
+                while ( ( exitStatus = exitStatusRaceConditionBugWorkaround(dtc, filePath,launcher) ) == null ) {
                     Thread.sleep(reCheckSleep);
                     byte[] newOutput;
                     try (InputStream is = stdoutTmp.readFromOffset( stdOut.size() )) {
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/OcWatch.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/OcWatch.java
@@ -92,6 +92,7 @@ public class OcWatch extends AbstractStepImpl {
             getContext().saveState();
             listener.getLogger().println( "Entering watch" );
 
+            Integer exitStatus = -1;
             try {
 
                 FilePath stderrTmp = filePath.createTextTempFile( "watchstderr", ".txt", "", false );
@@ -114,9 +115,6 @@ public class OcWatch extends AbstractStepImpl {
                         QuietTaskListenerFactory.QuietTasklistener quiet = QuietTaskListenerFactory.build(listener);
                         Controller dtc = task.launch(envVars,filePath,launcher,quiet);
 
-                        exitStatusRaceConditionBugWorkaround( dtc, filePath, launcher);
-
-                        Integer exitStatus = -1;
                         try {
                             long reCheckSleep = 250;
                             boolean firstPass = true;
@@ -153,16 +151,16 @@ public class OcWatch extends AbstractStepImpl {
                                 listener.getLogger().println( "Checking watch output again in " + reCheckSleep + "ms" );
                                 Thread.sleep( reCheckSleep );
 
-                            } while ( (exitStatus = dtc.exitStatus(filePath,launcher)) == null );
+                            } while ( (exitStatus = exitStatusRaceConditionBugWorkaround(dtc, filePath,launcher)) == null );
 
                         } finally {
                             dtc.cleanup(filePath);
                         }
 
-                        // Reaching this point means that the watch terminated
+                        // Reaching this point means that the watch terminated - exitStatus will not be null
                         if ( exitStatus.intValue() != 0 ) {
                             // Looks like the watch command encountered an error
-                            throw new AbortException( "watch terminated with an error: " + dtc.exitStatus(filePath,launcher) );
+                            throw new AbortException( "watch terminated with an error: " + exitStatus );
                         }
                     }
 


src/main/java/com/openshift/jenkins/plugins/pipeline/OcWatch.java
```diff
@@ -153,16 +151,16 @@ public Void run() {
                                 listener.getLogger().println( "Checking watch output again in " + reCheckSleep + "ms" );
                                 Thread.sleep( reCheckSleep );
 
-                            } while ( (exitStatus = dtc.exitStatus(filePath,launcher)) == null );
+                            } while ( (exitStatus = exitStatusRaceConditionBugWorkaround(dtc, filePath,launcher)) == null );
```
bparees: this could loop forever if something goes wrong and we can't get the exit code (existStatusRace... returns null after 30 retries), right?  I think we want a backstop.

src/main/java/com/openshift/jenkins/plugins/pipeline/OcAction.java
```diff
@@ -46,20 +46,21 @@ public OcAction(String server, String project, String verb, List verbArgs, List
...
+                return i; // exitStatus is healthy, we should be safe
             } catch ( IOException ioe ) {
                 if ( tries == 1 ) {
                     throw ioe;
                 }
-                Thread.sleep(1000);
+                Thread.sleep(125);
             }
         }
+        return null;
```
bparees: should log some details here when we give up, about what pid/file we were looking for and noting that we gave up on it.

gabemontero:
updates pushed ... not exactly what was asked for, but i believe it covers the intent

gabemontero:
cool thx

gabemontero:
Looks like an initial null in fact is possible:

```
Caught: java.io.IOException: dtc.exitStatus should return non-null integer or exception, but returned null; file path file:/var/lib/jenkins/jobs/extended-test-jenkins-pipeline-74m73-cj9j3-sample-pipeline-openshift-client-plugin/workspace/ and durable task org.jenkinsci.plugins.durabletask.BourneShellScript$ShellController@2267a4
[Pipeline] End of Pipeline
java.io.IOException: dtc.exitStatus should return non-null integer or exception, but returned null; file path file:/var/lib/jenkins/jobs/extended-test-jenkins-pipeline-74m73-cj9j3-sample-pipeline-openshift-client-plugin/workspace/ and durable task org.jenkinsci.plugins.durabletask.BourneShellScript$ShellController@2267a4
	at com.openshift.jenkins.plugins.pipeline.OcAction.exitStatusRaceConditionBugWorkaround(OcAction.java:56)
	at com.openshift.jenkins.plugins.pipeline.OcAction$Execution.run(OcAction.java:203)
	at com.openshift.jenkins.plugins.pipeline.OcAction$Execution.run(OcAction.java:141)
	at org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepExecution$1$1.call(AbstractSynchronousNonBlockingStepExecution.java:47)
	at hudson.security.ACL.impersonate(ACL.java:260)
	at org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepExecution$1.run(AbstractSynchronousNonBlockingStepExecution.java:44)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:748)
```

Going to restructure, retry at least 30 times before throwing exception on null.

gabemontero:
Yep, had it backwards ... it returns null until it exits in some fashion.

So this gets into long running tasks and how log to wait.  Given the best practice / preferred approach with pipeline scripts is to use `timeout(..) { ....}` and allow the script writer to "configure" timeouts in that fashion, generically, one could argue we shouldn't force arbitrary / additional timeouts in the logic here.

Going to revert the logic to the PR's original form, and confirm tests pass again.  We can then debate further as needed @bparees

gabemontero:
bump @bparees .... see https://github.com/openshift/jenkins-client-plugin/pull/29#issuecomment-305353125

bparees:
> Yep, had it backwards ... it returns null until it exits in some fashion.

?

my reading is that it loops up to 30 times trying to read the pid exit code.  if after 30 tries it hasn't done so successfully, it returns null, *unless* it is still getting an IOException(possible numberformatexception) on the 30th try, in which case it returns that exception.

so the `while(exitStatusWorkaround==null)` check can still loop forever if we are in a case where the exception is not an IOException (is filenotfound reported as an ioexception?), and we're still not logging any details in the case where we do give up and return null.

src/main/java/com/openshift/jenkins/plugins/pipeline/OcAction.java
gabemontero: log part of next push ... thought I had that in even after recent revert

gabemontero:
ok, first, yes, filenotfoundexception is an extension of ioexception

next, after looking again, I believe returning -1 is sufficient for both the steady state, the error scenario we are tracking here, and addresses your concern

updates will be pushed momentarily

gabemontero:
update pushed ... ptal when you get the chance

and we'll see how the pr job does

src/main/java/com/openshift/jenkins/plugins/pipeline/OcAction.java
```diff
@@ -195,11 +197,9 @@ protected OcActionResult run() throws Exception {
                 ByteArrayOutputStream stdErr = new ByteArrayOutputStream();
                 ByteArrayOutputStream stdOut = new ByteArrayOutputStream();
 
-                exitStatusRaceConditionBugWorkaround( dtc, filePath, launcher);
-
                 long reCheckSleep = 250;
                 Integer exitStatus;
-                while ( ( exitStatus = dtc.exitStatus(filePath,launcher) ) == null ) {
+                while ( ( exitStatus = exitStatusRaceConditionBugWorkaround(dtc, filePath,launcher) ) == null ) {
```
bparees: ok well now it's never going to be null so we'll never enter this loop....

src/main/java/com/openshift/jenkins/plugins/pipeline/OcWatch.java
bparees: and we'll never loop here (we'll always execute the do-while exactly once)

gabemontero:
On Mon, Jun 5, 2017 at 2:40 PM, Ben Parees <notifications@github.com> wrote:

> *@bparees* commented on this pull request.
>
> In src/main/java/com/openshift/jenkins/plugins/pipeline/OcAction.java
> <https://github.com/openshift/jenkins-client-plugin/pull/29#discussion_r120179335>
> :
>
> >                  long reCheckSleep = 250;
>                  Integer exitStatus;
> -                while ( ( exitStatus = dtc.exitStatus(filePath,launcher) ) == null ) {
> +                while ( ( exitStatus = exitStatusRaceConditionBugWorkaround(dtc, filePath,launcher) ) == null ) {
>
> ok well now it's never going to be null so we'll never enter this loop....
>
Incorrect.  Perhaps my comments on my earlier findings were unclear.  It
will return null until that task has finished.

Testing has confirmed this.

src/main/java/com/openshift/jenkins/plugins/pipeline/OcAction.java
gabemontero: To clarify, per my comment at https://github.com/openshift/jenkins-client-plugin/pull/29#issuecomment-305353125 .... `dtc.exitStatus(filePath,launcher)` returns null while the task is in progress and until that task finishes one way or the other

src/main/java/com/openshift/jenkins/plugins/pipeline/OcAction.java
```diff
@@ -46,20 +46,22 @@ public OcAction(String server, String project, String verb, List verbArgs, List
...
+    public static Integer exitStatusRaceConditionBugWorkaround( Controller dtc, FilePath filePath, Launcher launcher ) throws InterruptedException, IOException {
         for ( int tries = 30; tries > 0; tries-- ) {
             try {
                 // exitStatus can throw an IOException (reporting a NumberFormatException) if the PID file has been created but not
-                // populated. Make sure it stops throwing this exception before continuing.
-                dtc.exitStatus(filePath,launcher);
-                return; // exitStatus is healthy, we should be safe
+                // populated, or if the PID file has not yet been created. Make sure it stops throwing this exception before continuing.
+                Integer i = dtc.exitStatus(filePath,launcher);
+                return i; // exitStatus is healthy, we should be safe
```
bparees: why not just return dtc.exitStatus?  this is what confused me into thinking you'd never return null.

gabemontero:
On Mon, Jun 5, 2017 at 3:06 PM, Ben Parees <notifications@github.com> wrote:

> *@bparees* commented on this pull request.
>
> In src/main/java/com/openshift/jenkins/plugins/pipeline/OcAction.java
> <https://github.com/openshift/jenkins-client-plugin/pull/29#discussion_r120184998>
> :
>
> >          for ( int tries = 30; tries > 0; tries-- ) {
>              try {
>                  // exitStatus can throw an IOException (reporting a NumberFormatException) if the PID file has been created but not
> -                // populated. Make sure it stops throwing this exception before continuing.
> -                dtc.exitStatus(filePath,launcher);
> -                return; // exitStatus is healthy, we should be safe
> +                // populated, or if the PID file has not yet been created. Make sure it stops throwing this exception before continuing.
> +                Integer i = dtc.exitStatus(filePath,launcher);
> +                return i; // exitStatus is healthy, we should be safe
>
> why not just return dtc.exitStatus? this is what confused me into thinking
> you'd never return null.
>

Cause I didn't revert back that far after I realized what exitStatus was
doing :-) ... I'll add that change momentarily

gabemontero:
simply returning update pushed

bparees:
lgtm

gabemontero:
Cool thanks @bparees

PR #33, opened 2017-06-08
add some retry on watch

gabemontero:
@openshift/devex fyi

this is some more of the "retry" I've been referring to in scrum around the flakes we've been seeing in the pipeline jobs when attempting to access the openshift master via `oc` (the TLS handshake errors for example)

as I attempted to note in the comment around the change, this is intended to be a conservative/simplistic application of retry, where minus message parsing there isn't much to do around distinguishing error types

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/OcWatch.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/OcWatch.java
@@ -2,8 +2,10 @@ package com.openshift.jenkins.plugins.pipeline;
 
 import com.openshift.jenkins.plugins.util.ClientCommandBuilder;
 import com.openshift.jenkins.plugins.util.QuietTaskListenerFactory;
+
 import hudson.*;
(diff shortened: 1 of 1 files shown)


bparees:
lgtm

stevekuznetsov:
@gabemontero @bparees not sure this patch makes any sense. We're adding a retry to the _user_ _provided_ closure that gets run inside of the watch body. Not only does this mean that the `error` pipeline step gets eaten by this logic but also `timeout` will not work on watches. We should either add retry to the low-level `oc` actions or ask the user to do retries in their own code, but not this.

gabemontero:
OK, catching back up ...and I think I saw a PR that may have reverted this to at least some degree.  Saving a tab on this one as I step through my post holiday recovery, and will circle back, reconcile, officially chime in once I'm caught up.

gabemontero:
ok, I caught up to the PR around reverting this, etc.

I'll continue the discussion there.

PR #40, opened 2017-06-22
Don't use a human-readable printer for `oc run` invocation

stevekuznetsov:
A bug in the human readable printer implementation means that we need to
ask for a machine readable output from `oc run` in the sample pipeline
so that `oc` versions 3.6+ work correctly.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

/cc @gabemontero @mfojtik @bparees 

xref: https://github.com/openshift/origin/issues/14821

### DIFF
--- a/examples/jenkins-image-sample.groovy
+++ b/examples/jenkins-image-sample.groovy
@@ -139,6 +139,7 @@ try {
                 runargs1 << "jenkins-second-deployment"
                 runargs1 << "--image=docker.io/openshift/jenkins-2-centos7:latest"
                 runargs1 << "--dry-run"
+                runargs1 << "-o yaml"
                 openshift.run(runargs1)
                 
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
Whitespace diff... sorry -- let me fix that

bparees:
ok, @stevekuznetsov has convinced me of the dire need for this, so i'm ok to merge it.

gabemontero:
ok ... will hit button after test completes

i ran with -o yaml locally and confirmed it does work ... so expecting test to be OK

bparees:
@gabemontero after this goes in, you can revert my reversion of @stevekuznetsov's change.

gabemontero:
@stevekuznetsov has PR https://github.com/openshift/jenkins/pull/321 teed up for that as well

gabemontero:
following console logs ... test passed ... merging

PR #41, opened 2017-06-22
fix the -o yaml fix for oc run --dry-run test

gabemontero:
@bparees @stevekuznetsov FYI

we occasionally have a chicken / egg problem wrt testing our changes to the jenkins plugins; we can test plugin changes as part of PR testing, but we can't add new tests until the plugin version has been bumped and the jenkins image updated, since our overnight runs only pick up official images

also, if we change the test, the build config which references https://github.com/openshift/jenkins-client in our test case will pick off master branch and not the specific PR we are testing.

That happened here when I ran @stevekuznetsov test update locally and got an index out of bound error

Here are the changes I employed that allowed to test to run locally.


gabemontero:
I'll borrow the prhase "eventual consistency" to rationalize the situation :-)

Issue #37, opened 2017-06-21
Argument in wrong position when executing Selector.rollout().status("-w")


GolanTrevize10:
Hi, I tried to run the command Selector.rollout().status("-w") but it looks like the -w argument is placed in a wrong position and the oc command returns an error:

ERROR: rollout:status returned an error;
{reference={}, err=Error: unknown shorthand flag: 'w' in -w

Usage:
  oc rollout SUBCOMMAND [options]

Available Commands:
  cancel      cancel the in-progress deployment
  history     View rollout history
  latest      Start a new rollout for a deployment config with the latest state from its triggers
  pause       Mark the provided resource as paused
  resume      Resume a paused resource
  retry       Retry the latest failed rollout
  status      Show the status of the rollout
  undo        Undo a previous rollout

Use "oc <command> --help" for more information about a given command.
Use "oc options" for a list of global command-line options (applies to all commands)., verb=rollout, cmd=oc rollout -w status deploymentconfig/jws-app --insecure-skip-tls-verify --server=xxxxxx --namespace=pablodemo-test --token=XXXXX , out=, status=1}

gabemontero:
@GolanTrevize10 - similar to issue 36 if you like you can build a local copy of this plugin using the latest at https://github.com/openshift/jenkins-client-plugin to get the updated openshift.rollout().status("-w) behavior.

We most likely won't update the openshift jenkins centos images on docker hub until next week. I'll post an update here when we do.

But I am thinking of at least publishing a new version of the plugin at the jenkins update center later today if you want to wait for that.

thanks

GolanTrevize10:
Thanks, I tried it and it works

gabemontero:
Terrific - thanks for the update.

gabemontero:
final update: v0.9.6 has been cut and the images on docker hub should be updated to include them within the hour

PR #42, opened 2017-06-22
reverse userargs/verbargs to allow cmds like oc rollout status -w

gabemontero:
Fixes https://github.com/openshift/jenkins-client-plugin/issues/37

@bparees 

@jupierce - if you got a sec - my local tests are passing for this change, but I wanted to see if you recall some sort of scenario from your original testing that required verbArgs to be in front of userArgs for the ultimate `oc` invocation


jupierce:
@gabemontero I don't know of any downside there.

gabemontero:
cool, thanks for the confirmation @jupierce

PR #44, opened 2017-06-26
Put oc on PATH

gerbszt:
Unfortunately even, when declared inside 'tools' section in Declarative Pipeline, oc tool path doesn't show up on PATH and it needs to be added manually using withEnv expression:
```
pipeline {
    environment {
     OC_PATH = tool('oc1.5.1')
    }
    stages {
        stage("withEnv") {
            steps {
                script {
                    withEnv(['PATH+OC=$OC_PATH']) {
                        openshift.withCluster("cluster1") {
                            echo "${openshift.raw( "version" ).out}"
                        }
                    }
                }
            }
        }
   }
}
```
After applying the change in this PR it can be simplified to:
```
pipeline {
    tools {
        oc 'oc1.5.1'
    }
    stages {
        stage("withoutEnv") {
            steps {
                script {
                    openshift.withCluster("cluster1") {
                        echo "${openshift.raw( "version" ).out}"
                    }
                }
            }
        }
    }
}
```
It looks like saving two lines (and one level of indentation), but it's more then that. First, path to oc is no longer handled manually in the pipeline's code, which is supposed to be declarative, rather then imperative. Second, withEnv() doesn't have to be repeated over and over when openshift plugin is used in multiple stages.

This commit has been inspired by a similar method in [maven plugin](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/tasks/Maven.java#L535).

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/OpenShiftClientTools.java
+++ b/src/main/java/com/openshift/jenkins/plugins/OpenShiftClientTools.java
@@ -36,6 +36,13 @@ public class OpenShiftClientTools extends ToolInstallation implements Environmen
         return new OpenShiftClientTools(getName(), translateFor(node, log), getProperties().toList());
     }
 
+    @Override
+    public void buildEnvVars(EnvVars env) {
+        if (getHome() != null) {
+            env.put("PATH+OC", getHome());
+        }
+    }
+
     @Extension @Symbol("oc")
     public static class DescriptorImpl extends ToolDescriptor<OpenShiftClientTools> {
 


gabemontero:
[test]

gabemontero:
thanks again for the contribution @gerbszt !  good stuff - will merge upon test completion

gabemontero:
hmm ... the test run failed in an atypical spot (timed out creating objects from the mongodb template)

I'll do a sanity check locally with first master branch and then this PR's branch and see what transpires.

gabemontero:
yeah the pull runs for me locally just fine, as does master ... I think this PR should be fine, but to be safe
a) i've got another PR for a test addition unrelated to this change running right now
b) i'll start another test run for this PR in a sec

Based on those results, I'll merge this if the results reinforce for me that this change is unrelated to any failures.

gabemontero:
[test]

gabemontero:
Yeah, this test failure was the ruby build not getting scheduled and completed in time.

In discussion with @bparees our pr testing env probably has some unhealthiness to it (possibly disk related).  PR 45 saw odd issues as well.  Going to start debugging that separately.

Since this looks good in my local testing, merging.

Issue #55, opened 2017-09-08
start build with -F option doesn't print logs


GolanTrevize10:
I'm using plugin version 0.9.6 with Jenkins 2.74. 
$ oc version
oc v3.6.0+c4dd4cf
kubernetes v1.6.1+5115d708d7
features: Basic-Auth

Server xxxxxxxxxxxxx
openshift v3.6.173.0.5
kubernetes v1.6.1+5115d708d7

 
When I try to start build from Jenkins using the -F flag using this command
def bc = openshift.selector("bc/jws-app")
bc.startBuild('--follow') or bc.startBuild('-F)

Jenkins waits until build is finished but it does not print the logs in the console. If I start the build from command line
oc start-build bc/jws-app -F

then it works correctly and I can see the logs in the console

gabemontero:
Hi (again :-) ) @GolanTrevize10 

yeah, currently, this plugin's `startBuild` verb does not make the necessary changes in our interactions with the durable task plugin to stream the build logs in real time.

Short term, you can get the streaming behavior by removing the `--follow` and `-F` from your startBuild, and employ `bc.logs('-f')` .... see https://github.com/openshift/jenkins-client-plugin#actions-speak-louder-than-words

That said, @jupierce and I talked a bit, and allowing streaming logs to `startBuild` could be fairly clean and still in line with how we'd like to the fluent style syntax of this plugin to evolve.  I'll see if I can construct a quick patch in the next day or two for you to try; otherwise, I'll post a new requirement on our trello board.

@bparees fyi

gabemontero:
yep ... i've got something working ... pull the plugin from this zip file, install it, and see how this works for you  
[client-plugin.zip](https://github.com/openshift/jenkins-client-plugin/files/1293975/client-plugin.zip)

jupierce:
@GolanTrevize10 generally, all invocations of 'oc' by the plugin result in capturing the stdout and returning it to you in the resulting groovy object. This stdout/stderr is not generally streamed to the Jenkins console because it would be quite noisy. In this example:

`result = bc.startBuild('--follow')`

After invoking this, the 'result' variable would contain the stdout (along with other information like return code, command invoked, etc). You can see the fields of this return object in the console:
`echo "${result}"`

[...]

PR #56, opened 2017-09-11
mimic -F/--follow start-build behavior

gabemontero:
Fixes https://github.com/openshift/jenkins-client-plugin/issues/55

@openshift/devex @jupierce fyi

### DIFF
--- a/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
+++ b/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
@@ -11,6 +11,7 @@ import hudson.AbortException
 import hudson.FilePath
 import hudson.Util
 
+import java.util.Arrays;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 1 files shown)


gabemontero:
[test]

gabemontero:
bump @openshift/devex @jupierce

bparees:
lgtm

gabemontero:
[merge]

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -1141,13 +1142,19 @@ class OpenShiftDSL implements Serializable {
 
         public OpenShiftResourceSelector startBuild(Object... ouserArgs ) throws AbortException {
             String[] userArgs = toStringArray(ouserArgs);
+            List argList = Arrays.asList(userArgs);
+            boolean realTimeLogs = argList.contains("-F") || argList.contains("--follow=true") || argList.contains("--follow");
```
jupierce: if realTimeLogs, I would not call splitNames on the output. This will create a selector that tries to select each line of log output as if it were a named openshift resource. This is the primary reason folks shouldn't specify -F on startBuild (and should use .logs()) --  the selector it returns will not be of any use.

gabemontero: thanks for taking a look @jupierce 

First, to confirm, since we last talked and agreed to adding the -F / --follow support, you are still OK with the fact we added it, but you are noting some of the side effects, and clarifying that .logs() is preferred (which was my take from our original conversation).  Yes?

I'll open a separate pull shortly to address the `splitNames` piece.

Issue #64, opened 2017-10-24
add openshift.cluster() to output cluster name


coreydaley:
openshift.cluster() should return name of selected cluster (similar to openshift.project())

PR #65, opened 2017-10-24
Adds http(s) support to openshift.process and adds openshift.cluster()

### LINKED ISSUE #63: openshift.process should support http(s) urls
openshift.process should support http(s) urls by using the -f flag, similar to how openshift.(verb) works.

Code that needs to be fixed:
https://github.com/openshift/jenkins-client-plugin/blob/master/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy#L570-L604

Working example:
https://github.com/openshift/jenkins-client-plugin/blob/master/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy#L523-L556

coreydaley:
openshift.process can now accept an http/https url to a template
openshift.cluster returns the url of the cluster being used

Fixes #63 & #64

### DIFF
--- a/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
+++ b/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
@@ -249,6 +249,10 @@ class OpenShiftDSL implements Serializable {
         return currentContext.getProject();
     }
 
+    public String cluster() {
+        return currentContext.getServerUrl();
+    }
+
     /**
      * @param name The name can be a literal URL for the clusterName or,
      *          preferably, a Jenkins specific
@@ -527,9 +531,8 @@ class OpenShiftDSL implements Serializable {
             obj = JsonOutput.toJson( obj );
         }
 
-        boolean markup = false;
         String s = obj.toString();
-        markup = s.contains("{") || s.contains(":"); // does this look like json or yaml?
+        boolean markup = s.contains("{") || s.contains(":"); // does this look like json or yaml?
         boolean httpref = s.toLowerCase().startsWith("http") && verb.equalsIgnoreCase("create"); // a create from a http or https raw.githubuser path
 
         Result r = new Result( verb );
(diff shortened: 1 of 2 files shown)


coreydaley:

coreydaley:
[test]

coreydaley:
@gabemontero 
Addressed item 1 in above comment, added some openshift.cluster() usage to examples

gabemontero:
I'll let the test complete, then merge.

Probably ought to wait on 3.7 shipping before cutting a new version of the plugin at this point, but we can discuss that with @bparees either here or in scrum

gabemontero:
[merge]

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -582,21 +584,24 @@ class OpenShiftDSL implements Serializable {
...
+        if ( markup && !httpref ) { // does this look like json or yaml?
+            FilePath f = currentContext.exec.getWorkspaceFilePath().createTextTempFile( "process", ".markup", s, false );
             try {
-                r = new Result( "process" );
                 r.actions.add((OcAction.OcActionResult)script._OcAction( buildCommonArgs("process", ["-f", f.getRemote() ], args, "-o=json" ) ));
                 r.failIf( "process returned an error" );
             } finally {
                 f.delete();
             }
+        } else if (httpref) {
```
bparees: if it starts with a { why would we ever treat it as a http url?

bparees: ok i guess i see how the logic works.  looks fragile but i guess this isn't making it worse.

bparees:
> Probably ought to wait on 3.7 shipping before cutting a new version of the plugin at this point, but we can discuss that with @bparees either here or in scrum

this seems like a bug people are going to hit, i'd rather fix it before we ship.

gabemontero:
10-4 I'll start the update process

gabemontero:
awaiting v1.0.1 at https://updates.jenkins-ci.org/download/plugins/openshift-client/

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
coreydaley: I tried to duplicate what was there, i plan on doing some further refactoring but didn't want to mix that in this pr

gabemontero:
centos image update https://github.com/openshift/jenkins/pull/408

gabemontero:
rhel image update https://buildvm.openshift.eng.bos.redhat.com:8443/job/devex/job/devex%252Fjenkins-plugins/8/console

Issue #79, opened 2017-11-15
Add support for "oc rsh"


elconas:
It seems that "oc rsh" is not working with the```openshift.raw()``` interface of the plugin: 

Pipeline:

```
node {
  def ocDir = tool "oc"
  withEnv(["PATH+OC=${ocDir}"]) {
  openshift.withCluster('minishift') {
    openshift.withProject('myproject') {
      echo "HELLO FROM ${openshift.project()}"
      echo openshift.raw('version').out
      echo openshift.raw('get','pod', 'petclinic2-3-hjx72').out
      echo openshift.raw("rsh","petclinic2-3-hjx72","hostname").out
    }
  }
  }
}
```

I get the following:

```
Started by user admin
[Pipeline] node
Running on Jenkins in /var/lib/jenkins/workspace/openshift-test
[Pipeline] {
[Pipeline] tool
[Pipeline] withEnv
[Pipeline] {
[Pipeline] echo

[Pipeline] _OcContextInit
[Pipeline] _OcContextInit
[Pipeline] echo
HELLO FROM myproject
[Pipeline] _OcAction
[Pipeline] echo
oc v3.6.0+c4dd4cf
kubernetes v1.6.1+5115d708d7
features: Basic-Auth GSSAPI Kerberos SPNEGO

Server https://xxx.xxx.xxx.xxx:8443/
openshift v3.6.0+c4dd4cf
kubernetes v1.6.1+5115d708d7

[Pipeline] _OcAction
[Pipeline] echo
NAME                 READY     STATUS    RESTARTS   AGE
petclinic2-3-hjx72   1/1       Running   4          39d

[Pipeline] _OcAction
[Pipeline] }
[Pipeline] // withEnv
[Pipeline] }
[Pipeline] // node
[Pipeline] End of Pipeline
ERROR: raw command [rsh, petclinic2-3-hjx72, hostname] returned an error;
{reference={}, err=Error from server (NotFound): pods "petclinic2-3-hjx72" not found, verb=, cmd=oc  rsh petclinic2-3-hjx72 hostname --insecure-skip-tls-verify --server=https://xxx.xxx.xxx.xxx:8443/ --namespace=myproject --token=XXXXX , out=, status=1}

Finished: FAILURE
```

gabemontero:
[...]

PR #81, opened 2017-11-16
add first class rsh, better deal with args for proxy cmds like exec, rsh

gabemontero:
Fixes https://github.com/openshift/jenkins-client-plugin/issues/79

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/util/ClientCommandBuilder.java
+++ b/src/main/java/com/openshift/jenkins/plugins/util/ClientCommandBuilder.java
@@ -71,14 +71,11 @@ public class ClientCommandBuilder implements Serializable {
 
         String toolName = (new OpenShift.DescriptorImpl()).getClientToolName();
         cmd.add(toolName);
-        cmd.add(verb);
-
-        cmd.addAll(toStringArray(verbArgs));
-
-        cmd.addAll(toStringArray(userArgs));
-
-        cmd.addAll(toStringArray(options));
-
+        
+        // in general with 'oc' having arguments like --server or --namespace precede the verb helps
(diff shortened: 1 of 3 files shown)


gabemontero:
[test]

src/main/java/com/openshift/jenkins/plugins/util/ClientCommandBuilder.java
```diff
@@ -111,6 +106,14 @@ private boolean hasArg(List<String> args, String... argsToFind) {
...
             }
         }
+        
+        cmd.add(verb);
+
+        cmd.addAll(toStringArray(verbArgs));
+
+        cmd.addAll(toStringArray(userArgs));
+
+        cmd.addAll(toStringArray(options));
```
bparees: this is just ensuring that the "--" arg does not come before the "--loglevel" or other args being added?

gabemontero: ah ... i should update the comment, forgot that the literal `--` sequence could be a valid arg

what I was trying to convey was that things like --server, --namespace, --token, --loglevel in the wrong position would confuse `oc rsh` ... that is what was causing the problem .... moving those args prior to the verb works generically, and protects `oc rsh`

bparees: got it, yeah a clarifying comment would help, otherwise lgtm.

gabemontero: comment clarified

gabemontero:
[merge]

PR #86, opened 2017-11-17
Fixes order of flags when using rsh

### LINKED ISSUE #82: new oc rsh parially broken
Hello, I tested the code introduced by https://github.com/openshift/jenkins-client-plugin/pull/81 and ran into an issue

```
run -it --rm -v $(pwd):/jenkins-plugin -v $(pwd)/../.m2:/root/.m2 -w /jenkins-plugin maven:latest mvn package
```
Installed the plugin, however some commands still fail:

Pipeline Code:
```
  echo openshift.rsh(shortname,"ps ax").out
```
Result:

```
ERROR: rsh returned an error;
{reference={}, err=error: unknown gnu long option

Usage:
 ps [options]

 Try 'ps --help <simple|list|output|threads|misc|all>'
  or 'ps --help <s|l|o|t|m|a>'
 for additional help text.

For more details see ps(1).
command terminated with exit code 1, verb=rsh, cmd=oc --server=https://xxx.xxx.xxx.xxx:8443/ --namespace=myproject --token=XXXXX rsh centos7-im0t0k1k-1-62s9v ps ax --insecure-skip-tls-verify , out=, status=1}
```
It seems you you have to move ALL --param params in front of the command.

This works:

Pipeline:
```
echo openshift.rsh(shortname,"echo").out
```
Result:
```
[Pipeline] _OcAction
[Pipeline] echo
--insecure-skip-tls-verify
```

coreydaley:
Corrects the order of the flags in buildCommonArgs to place
--insecure-skip-tls-verify before the userArgs

Fixes #82

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/ClusterConfig.java
+++ b/src/main/java/com/openshift/jenkins/plugins/ClusterConfig.java
@@ -63,12 +63,12 @@ public class ClusterConfig extends AbstractDescribableImpl<ClusterConfig>
     }
 
     public boolean isSkipTlsVerify() {
-        return skipTlsVerify;
+        return this.skipTlsVerify;
     }
 
     @DataBoundSetter
-    public void setSkipTlsVerify(boolean skipTlsVerify) {
-        this.skipTlsVerify = skipTlsVerify;
+    public void setSkipTlsVerify(boolean skipTLSVerify) {
+        this.skipTlsVerify = skipTLSVerify;
     }
 
     public String getDefaultProject() {
--- a/src/main/java/com/openshift/jenkins/plugins/freestyle/BaseStep.java
+++ b/src/main/java/com/openshift/jenkins/plugins/freestyle/BaseStep.java
@@ -61,7 +61,7 @@ public abstract class BaseStep extends Builder {
     public String getClusterName() {
         return clusterName;
     }
-    
+
     public String getClusterName(Map<String, String> overrides) {
         return getOverride(getClusterName(), overrides);
     }
@@ -132,28 +132,31 @@ public abstract class BaseStep extends Builder {
     protected boolean runOcCommand(final AbstractBuild build,
             final TaskListener listener, final String verb,
             final List verbArgs, final List userArgs, final List options,
-            final List verboseOptions, final OcProcessRunner runner)
+            final OcProcessRunner runner)
             throws IOException, InterruptedException {
         final Map<String, String> overrides = consolidateEnvVars(listener, build, null);
         ClusterConfig c = getCluster(overrides);
         final String server, project, token, caContent;
+        String selectedCAPath = "";
+        boolean shouldSkipTLSVerify = false;
+
+        ArrayList<String> advArgs = new ArrayList<String>();
 
         if (advancedArguments != null) {
-            for (AdvancedArgument aa : advancedArguments) {
-                userArgs.add(aa.getValue(overrides));
+            for (AdvancedArgument advArg : advancedArguments) {
+                advArgs.add(advArg.getValue(overrides));
             }
         }
 
         if (c == null) { // if null, we assume the cluster is running the
                          // Jenkins node.
             server = ClusterConfig.getHostClusterApiServerUrl();
-            verboseOptions.add("--certificate-authority="
-                    + SERVICE_ACCOUNT_CA_PATH);
+            selectedCAPath = SERVICE_ACCOUNT_CA_PATH;
             caContent = null;
         } else {
             server = c.getServerUrl();
             if (c.isSkipTlsVerify()) {
-                options.add("--insecure-skip-tls-verify");
+                shouldSkipTLSVerify = true;
                 caContent = null;
             } else {
                 caContent = c.getServerCertificateAuthority();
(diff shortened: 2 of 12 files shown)


coreydaley:
[test]

coreydaley:
@gabemontero ptal

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -384,8 +384,8 @@ class OpenShiftDSL implements Serializable {
                 project:(getProject ? currentContext.getProject() : null),
                 verb:verb,
                 verbArgs:verbArgs,
-                userArgs:userArgsList,
                 options:optionsBase,
+                userArgs:userArgsList,
```
gabemontero: @coreydaley - are you sure this reconciles with the `ClientCommandBuilder` constructor?

There, the order is verbArgs, options, userArgs ...

It might better to pass the skip tls verification as a separate param just like server, token, project, and then adjust ClientCommandBuilder.buildCommand to prepend like we did with those

I'm also curious about the server certificate auth option --certificat-authority ... it is added to verboseOptionsBase, but then in ClientCommandBuilder.buildCommand ... that command is only added if !redacted or loglevel is greater than 0 ... I would think we would want to add that regardless, but perhaps not redact based on redacted/loglevel.  Please look into that as well.

src/main/java/com/openshift/jenkins/plugins/util/ClientCommandBuilder.java
```diff
@@ -91,12 +101,12 @@ private boolean hasArg(List<String> args, String... argsToFind) {
...
-            if (!hasArg(cmd, "--loglevel")) {
-                cmd.add("--loglevel=" + logLevel);
-            }
-        }
+        // if (!redacted || logLevel > 0) {
+        //     cmd.addAll(toStringArray(verboseOptions));
+        //     if (!hasArg(cmd, "--loglevel")) {
+        //         cmd.add("--loglevel=" + logLevel);
+        //     }
+        // }
```
gabemontero: I could see swizzling this some because of what you did with `verboseOptions` (i.e. the cmd.addAll call), but completing removing the setting of `--loglevel` is not something we should remove, and I'm not seeing where it is set elsewhere here

And the use of asString(boolean) where the parameter is try still exists wrt the redacted var/param

coreydaley: Good catch, forgot i left that commented out, fixed in next push

src/main/java/com/openshift/jenkins/plugins/freestyle/BaseStep.java
```diff
@@ -236,11 +237,11 @@ public boolean perform(String filename) throws IOException,
     }
 
     protected boolean standardRunOcCommand(final AbstractBuild build,
-            final TaskListener listener, String verb, List verbArgs,
-            List userArgs, List options, List verboseOptions)
+            final TaskListener listener, boolean skipTLSVerify, String caPath, String verb, List verbArgs,
```
gabemontero: Missed this last time, but noticed it this time, and then I checked out the branch and searched in eclipse to be sure .... for every call to this method, skipTLSVerify is false and caPath is ""

I think you need to parse the "Advanced Arguments" for the associated `oc` arguments and then set these parameters accordingly.

coreydaley:
@gabemontero comments addressed, ptal
Also, i renamed isSkipTlsVerify to isSkipTLSVerify, just FYI as you are looking over the PR

src/main/java/com/openshift/jenkins/plugins/freestyle/BaseStep.java
```diff
@@ -147,13 +149,12 @@ protected boolean runOcCommand(final AbstractBuild build,
...
             server = ClusterConfig.getHostClusterApiServerUrl();
-            verboseOptions.add("--certificate-authority="
-                    + SERVICE_ACCOUNT_CA_PATH);
+            selectedCAPath = SERVICE_ACCOUNT_CA_PATH;
             caContent = null;
         } else {
             server = c.getServerUrl();
-            if (c.isSkipTlsVerify()) {
-                options.add("--insecure-skip-tls-verify");
+            if (c.isSkipTLSVerify()) {
```
gabemontero: OK, I see where you've removed the skip tls / ca cert params that were being passed in, but whose values were always false and "", but I'm still concerned about the advance arguments.  

I realize now my prior comment about parsing them was not entirely sufficient.  It is not just that we need to parse for the skip tls and certs arg specifically.  In case of the "raw" command, I realize now when looking at the jelly files for the config under src/main/resources that the advance args from base step apply specifically to the  options of `oc` that start with `--` that work across all commands.  The raw step has a separate field that applies to the command being run. And as such, the advance args should appear prior to the invocation of the command in question.

The code and its currently lumping of advanced args into userArgs then muddles things when we order things in OcAction, especially in the case of rsh/exec.

We may have to add new config fields.

A blue jeans chat where we look at the UI panels, review the various oc options, etc. might be warranted wrt how we move forward.

Now, the create/delete are less clear

gabemontero: At worst, maintaining a specific set of options might be needed (though certainly it is desirable to avoid as options will be added, etc. over time)

src/main/java/com/openshift/jenkins/plugins/ClusterConfig.java
```diff
@@ -62,13 +62,13 @@ public void setServerCertificateAuthority(String serverCertificateAuthority) {
...
-    public boolean isSkipTlsVerify() {
-        return skipTlsVerify;
+    public boolean isSkipTLSVerify() {
+        return skipTLSVerify;
     }
 
     @DataBoundSetter
-    public void setSkipTlsVerify(boolean skipTlsVerify) {
-        this.skipTlsVerify = skipTlsVerify;
+    public void setSkipTLSVerify(boolean skipTLSVerify) {
```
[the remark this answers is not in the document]
coreydaley: The only other instance of Tls in the project is in main/resources/com/openshift/jenkins/plugins/ClusterConfig/config.jelly and it will be updated to TLS in the next push, good catch, thanks!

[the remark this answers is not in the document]
coreydaley: I just want to confirm, you mean setting skipTlsVerify/skipTLSVerify via pipeline code in a job correct? Such as using the openshift.setServerUrl() ?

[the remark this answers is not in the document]
gabemontero: well, it can be either freestyle or pipeline jobs wrt cluster config ... both type of jobs leverage them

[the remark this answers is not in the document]
coreydaley: This has been updated in the current push, i had to revert some of the TLS to Tls for backward compatibility

coreydaley:
@gabemontero ptal, feedback has been addressed including what we talked about

coreydaley:
@gabemontero commits have been squashed

gabemontero:
[merge]

Issue #89, opened 2017-11-29
Enable CA Verify with insecure-skip-tls-verify: true is broken


elconas:
When having this configured in ~jenkins/.kube/config on the os level
 
```
- cluster:
    insecure-skip-tls-verify: true
```
 
I can not enable SSL verify on the cluster Level in Jenkins. I will always get:
 
```
  action failed: {reference={}, err=error: specifying a root certificates file with the insecure flag is not allowed, verb=get
```
 
This should be possible. The plugin should not depend on System configuration

  * Setting KUBECONFIG to something sane, or 
  * overwriting insecure-skip-tls-verify always (even if false)) or (best)
  * allow setting ENV VARS at the global cluster level.
 
Workaround: Wrap in withEnv(["KUBECONFIG=${workspace}"]) - with deleteDir() in a finally block

gabemontero:
hey @elconas  - "depend" is not really the right word in "The plugin should not depend on System configuration" ... I would say ""... should not get tripped up ... "  ;-)

this issue is the plugin to date has been assuming there is no `.kube/config` file present, hence when
the oc binary is forked and finds the `.kube/config` path off of `$HOME`, it reads in those settings and they conflict with what the args to `oc` that the plugin constructs prior to forking the `oc` command.

so both short and long term possibilities exist wrt to precedence of config specified, and whether the plugin should even support the use of `.kube\config`, and if so, orders of precedence choices and option merging.

I'm currently of the following opinion
1) short term, do not support use of `.kube/config1, and essentially mimic @elconas 's workaround - set `KUBECONFIG` to a location that does not have the conflicting `.kube/config` file

In some experimentation bring up our jenkins image via the template in a pod, I am seeing that KUBECONFIG=/var/lib/origin/openshift.local.config/master/admin.kubeconfig is set.  However, that file does not in fact exist in the pod.  I'm leaning toward that (vs. WORKSPACE, where a person running jenkins outside of an openshift pod might leverage that location, even accidentally, for `oc login`.

And drive this change via this issue and the PR I'll compose for it.

2) long term, open a trello feature card for exploring the requirement for devising a config merging order of precedence similar to the `oc` binary, but include any arg overrides introduced via use of this plugin's DSL as part of that precedence. 

@bparees @coreydaley @elconas - opinions of course welcome

thanks

bparees:
> KUBECONFIG=/var/lib/origin/openshift.local.config/master/admin.kubeconfig

[...]

PR #90, opened 2017-11-29
protect from presence of ad hoc .kube/config files

gabemontero:
Fixes https://github.com/openshift/jenkins-client-plugin/issues/89

@elconas - if you think you would have the cycles, I can attach an hpi file with this change for you to try prior to merging / cutting a new version of the plugin - thanks

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/pipeline/OcAction.java
+++ b/src/main/java/com/openshift/jenkins/plugins/pipeline/OcAction.java
@@ -208,6 +208,10 @@ public class OcAction extends AbstractStepImpl {
                         .println(
                                 "Consider removing those options from startBuild and using the logs() command to follow the build output.");
             }
+            
+            // explicitly set KUBECONFIG such that use of this plugin in a non-openshift pod will not pick up a rogue .kube/config
+            // when we fork `oc`; for now, using the default setting when running in a openshift pod (presuming that is safe enough)
+            envVars.put("KUBECONFIG", "/var/lib/origin/openshift.local.config/master/admin.kubeconfig");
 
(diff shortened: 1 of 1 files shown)


gabemontero:
[test]

gabemontero:
there was a problem downloading a maven dependency during the build stage of the pr test

i'll give upstream a bit to settle, and re-kick

bparees:
I have some concerns about removing a feature which is that users can set KUBECONFIG to point to the config they want to use.

i'd also like to better understand where the existing KUBECONFIG env var is coming from before we go stomping it.

I don't think this problem is so urgent that we need to push a fix in w/o considering the other items you mentioned in the issue.

gabemontero:
yep, fair enough - as much as anything, I wanted to generate the discussion

I also realized, though didn't get around to addressing, that the plugin change could do things like 
a) see if some likely locations for `.kube/config` like JENKINS_HOME or WORKSPACE existed before overriding KUBECONFIG
b) see if the new default is not in place for KUBECONFIG prior to running

And I realized a few moments ago
c) and that the change in the jenkins image to base off of the openshift/origin image was in the 3.7 timeframe; so that explains why @elconas was not seeing `KUBECONFIG` set and had to explicitly set it ... in theory, if he moves to a 3.7 image, it will be set to the default from the openshift/origin anyway, and bypass the local `.kube/config` file he had.

gabemontero:
I'm going to close this pull and the associated issue out based on my point c) from https://github.com/openshift/jenkins-client-plugin/pull/90#issuecomment-348304477 about the 3.7 image change,
along with the other concerns noted

PR #107, opened 2018-01-05
Upd tests

gabemontero:
@openshift/sig-developer-experience @jupierce @stevekuznetsov fyi / ptal

So this enables some tests for recent changes submitted by @jupierce, @coreydaley , and myself now that a version of the plugin has been cut and the openshift jenkins centos image has been updated.

However, in some manual testing, I discovered some additional errors with narrow() and related() that occurred when I both a) added the new tests and ran the test pipeline once, as well as b) ran our test job a second time.  There are actually some different paths exercised when the script creates object vs. finds existing object.

And I realized that our current PR tester / overnight run only runs the test pipeline once.

So the roadmap will be:
1) verify this PR passes in ci
2) since there are fixes involved to make this pass, will have to comment out the test changes before merging
3) cut a new version of the plugin, update the image
4) (re)enable these tests
5) lastly, update the extended test in openshift/origin to run this pipeline twice

### DIFF
--- a/examples/jenkins-image-sample.groovy
+++ b/examples/jenkins-image-sample.groovy
@@ -21,6 +21,7 @@ try {
                 def templateExists = templateSelector.exists()
 
                 def templateGeneratedSelector = openshift.selector('all,secret', [ template: 'mongodb-ephemeral-template' ])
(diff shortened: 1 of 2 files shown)


gabemontero:
[test]

gabemontero:
OK, step 1 from the description was successful;  just pushed changes for step 2 and squashed; assuming passing tests, will merge

Issue #104, opened 2017-12-27
NPE when missing object is not helpful to users


stevekuznetsov:
Whenever I forget to wrap a call in `withCluster()` or try to `.object()` on something that doesn't exist, I get:

```
java.lang.NullPointerException: Cannot invoke method isSkipTlsVerify() on null object
	at org.codehaus.groovy.runtime.NullObject.invokeMethod(NullObject.java:91)
	at org.codehaus.groovy.runtime.callsite.PogoMetaClassSite.call(PogoMetaClassSite.java:48)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:48)
	at org.codehaus.groovy.runtime.callsite.NullCallSite.call(NullCallSite.java:35)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:48)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:113)
	at com.cloudbees.groovy.cps.sandbox.DefaultInvoker.methodCall(DefaultInvoker.java:19)
	at com.openshift.jenkins.plugins.OpenShiftDSL.buildCommonArgs(jar:file:/var/lib/jenkins/plugins/openshift-client/WEB-INF/lib/openshift-client.jar!/com/openshift/jenkins/plugins/OpenShiftDSL.groovy:368)
	at com.openshift.jenkins.plugins.OpenShiftDSL.buildCommonArgs(jar:file:/var/lib/jenkins/plugins/openshift-client/WEB-INF/lib/openshift-client.jar!/com/openshift/jenkins/plugins/OpenShiftDSL.groovy:351)
	at com.openshift.jenkins.plugins.OpenShiftDSL$OpenShiftResourceSelector._asSingleMap(jar:file:/var/lib/jenkins/plugins/openshift-client/WEB-INF/lib/openshift-client.jar!/com/openshift/jenkins/plugins/OpenShiftDSL.groovy:1069)
	at com.openshift.jenkins.plugins.OpenShiftDSL$OpenShiftResourceSelector.object(jar:file:/var/lib/jenkins/plugins/openshift-client/WEB-INF/lib/openshift-client.jar!/com/openshift/jenkins/plugins/OpenShiftDSL.groovy:1085)
```

[...]

PR #108, opened 2018-01-08
generic safety check on currentContext

gabemontero:
@jupierce @stevekuznetsov - update per last few exchanges in https://github.com/openshift/jenkins-client-plugin/issues/104

### DIFF
--- a/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
+++ b/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
@@ -243,6 +243,16 @@ class OpenShiftDSL implements Serializable {
             }
         }
     }
+    
+    @NonCPS
+    // for checks in private methods where the public method used, to included in the exception message, is unclear
+    // or embedded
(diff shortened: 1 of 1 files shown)


gabemontero:
[test]

stevekuznetsov:
So the error message is now `"proper use of openshift.withCluster() could not be detected"`?

It might be a bit clearer to and end user to say something to make it clear that it is improper to execute whatever thing the user is trying to do outside of a `withCluster()` call -- it's not immediately clear what a "proper" use is. Or maybe that's too specific? As a dumb user I just want to hear "hey, dumbo, you typed `openshift.selector()` outside of an `openshift.withCluster()` block. That's illegal."

gabemontero:
sure thing message updated

stevekuznetsov:
Thanks @gabemontero

gabemontero:
On Tue, Jan 9, 2018 at 12:33 PM Steve Kuznetsov <notifications@github.com>
wrote:

> Thanks @gabemontero <https://github.com/gabemontero>
>
np thanks for the prodding

Issue #96, opened 2017-12-19
'incorrect namespace' error from create() with objects for other namespaces


alanbchristie:
_This issue was recently raised on the OpenShift RedHat Email List and recreated here for comment/consideration._

I have a template that can be successfully processed and the objects created using oc from the command-line. The template is supposed to run in one namespace (let’s call it **Y**) but it creates secrets that are placed in another namespace/project (let’s call that **X**). The namespaces are managed the same user. Both namespaces exist and the following command, when run on the command-line, is valid and produces the expected results:

`oc process -f <template-file> | oc create -f -`

But you cannot do this in a Jenkins pipeline without getting an error. Here's my pipeline code...

```
openshift.withCluster("${CLUSTER}") {
    openshift.withProject(“${Y}") {
        def objs = openshift.process('—filename=<template-file>’)
        openshift.create(objs)
    }
}
```

When I do this I get the following error reported in the Jenkins Job output:

`err=error: the namespace from the provided object “X" does not match the namespace “Y". You must pass '—namespace=X' to perform this operation., verb=create`

How do you replicate the actions that appear to be legitimate from the command-line by using the pipeline plugin? The plugin appears to assume that the objects passed to `create()` must reside in the project namespace, but they do not have to.

Incidentally, I can work-around the problem by iterating through the list of objects created by the call to `process()`, i.e. by doing this...

```
def objs = openshift.process('—filename=<template-file>’)
for (obj in objs) {
    if (obj.metadata.namespace == “X") {
        openshift.create(obj, "—namespace=X")
    } else {
        openshift.create(obj)
    }
}
```

Shouldn't `create()`, like its command-line-counterpart, honour the object namespace?

gabemontero:
@alanbchristie yep, at face value I would agree.

Ultimately, we are talking about some rework of https://github.com/openshift/jenkins-client-plugin/blob/master/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy#L497-L528
where we possibly need to make more RPCs, and we have a bit of regression testing to perform
since that code path is common to a lot of our oc verb equivalents.

Because of the RPC implications, I'm wondering if we want to introduce and optimization option which would allow the code to assume a common namespace.

As such, I'd like to have some discussion on our (openshif dev) end to get to a consensus on things.  But certainly chime in with any thoughts along those lines you have as well @alanbchristie 

@openshift/sig-developer-experience fyi / ptal

[...]

PR #113, opened 2018-01-17
need to account for multiple projects, or projects outside the scoped…

gabemontero:
… project, for api object manipulation

Fixes https://github.com/openshift/jenkins-client-plugin/issues/96

### DIFF
--- a/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
+++ b/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
@@ -266,15 +266,24 @@ class OpenShiftDSL implements Serializable {
     }
 
     public String project() {
-        return currentContext.getProject();
+        if (currentContext != null)
+            return currentContext.getProject();
+        else
+            return null; 
     }
 
     public String cluster() {
-        return currentContext.getServerUrl();
+        if (currentContext != null)
+            return currentContext.getServerUrl();
+        else
+            return null; 
     }
 
     public boolean skipTLSVerify() {
-        return currentContext.isSkipTLSVerify();
+        if (currentContext != null)
+            return currentContext.isSkipTLSVerify();
+        else
+            return false; 
     }
 
     /**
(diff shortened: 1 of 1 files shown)


gabemontero:
check that - need to rebase

gabemontero:
ok now it is good to look at

gabemontero:
[test]

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -525,40 +534,127 @@ class OpenShiftDSL implements Serializable {
...
+                        // consistency check
+                        if (!found) {
+                            found = true;
+                            namespace = ns;
+                        } else {
+                            if (namespace == null && ns == null)
+                                continue;
+                            if (namespace != null && namespace.equals(ns))
+                                continue;
+                            if (ns != null && ns.equals(namespace))
```
bparees: this can only be true if the previous check was also true, so it's redundant, right?  (Unless someone screwed up the reflexive logic on an equals implementation)

gabemontero:
As to adding a test for multi-project and non-default project object list creation and viewing, I'll first need to augment the extended tests to create a second project and give the service account edit access to it.

Once there, the pipelines / templates that leverage the new project I used for my manual testing can be added and called from the extended tests.

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
gabemontero: yep the previous 2 checks make this redundant ... I'll update

gabemontero:
update from @bparees nit pushed

gabemontero:
some NPEs showed up with the complete suite when run on the ci job ... hopefully should have updates shortly

gabemontero:
[test]

gabemontero:
i'm having some trouble accessing rubygems locally, but the PR test job is looking better with the new commit

let's see how the ci job does

gabemontero:
ok, first, i hate groovy within jenkins pipelines (cps, noncps, constructors, reading files ... blllaaaahhhh !!) :-)

next, just pushed the update that locally for me passes:
- running our existing test job twice in row (a la what we do in PR testing and overnight runs)
- process template / create objs / access objs with multiple namespaces 
- newApp template / access objs with multiple namespaces 

the last two are new extended tests we'll want to add to the suite once we get the new client plugin baked;
and would love to figure out how to run the test pipeline specified in the PR we are using (seems like we should be able, just need to grok all the ci job / ext test / bc cfg / pr branch pieces needed)

gabemontero:
ok great ... the new commit is ready for review @openshift/sig-developer-experience as time permits for y'all

bparees:
i'd like to suggest this be held until post 3.9 given the extent of the changes.

gabemontero:
I was thinking that too.  Or at least until we cut the 3.9 branch for the jenkins repo.  Based on demand, we'll see if an update to the jenkins update center can help some folks in the interim, which is not pulled into the openshift jenkins image until post 3.9.

I'll update the issue with this.

bparees:
looks cleaner this iteration.  lgtm w/ retaining the "not for 3.9" caveat recommendation.

gabemontero:
gcs credential upload failed

/retest

PR #125, opened 2018-03-07
Added add and remove methods to descriptor to allow easier groovy config

garethahealy:
Added add / remove methods to descriptor
Added example groovy script in examples folder

https://trello.com/c/40z9fXkl/1413-3-provide-example-on-how-to-configure-the-plugin-via-groovy-jenkinsintegration

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ Otherwise, if you are interested in building this plugin locally, follow these s
 3. In the root of the local repository, run maven
     ```
     cd jenkins-client-plugin
-    mvn
+    mvn clean package
     ```
 4. Maven will build target/openshift-client.hpi  (the Jenkins plugin binary)
 5. Open Jenkins in your browser, and navigate (as an administrator):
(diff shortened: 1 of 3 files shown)


gabemontero:
[test]

gabemontero:
Awesome @garethahealy - thanks!

Assuming this passes our regression test, I'll go ahead and merge.

- update the readme to make note is this capability
- add an automated test, which would most likely include extending the jenkins image to include something like  examples/configure-jenkins-client.groovy, but then launch that image and confirm we can launch a client plugin pipeline in a pod running that image that leverages a `withCluster('cluster config from groovy file')`

@bparees @coreydaley fyi ^^

gabemontero:
note wrt automated tests ... our test groovy could read in all the cluster mounted information in the pod, just like the plugin does to populate the *new* cluster def, but name the cluster something else and then have the pipeline leverage the new name in its `withCluster` call.

gabemontero:
v1.0.8 has been cut and https://github.com/openshift/jenkins/pull/543 should update the openshift jenkins centos image when it merges

this will be slated for the v3.10 openshift jenkins rhel image (as 3.9 will ship this month, current 3.10 projection is this fall .... september/october)

Issue #131, opened 2018-03-21
version 1.0.8 openshift.selector doesn't work anymore


ramato-procon:
From version 1.0.8 the openshift.selector with pod and label doesn't work anymore.
Method-Call: 
```
openshift.withCluster(crp) {
    openshift.withProject(ocstage) {
       openshift.selector("pod", [ app: "${app}" ]) 
    }
}
```

Result:
`ERROR: projects were not set for selector([name=pod],[labels={app=mq-synchronizer}],[namelist=null],[projectlist=null])`

With version 1.0.7 it runs perfectly.

gabemontero:
hey @ramato-procon - thanks for the report

the `projects were not set for selector` message come from `selector.withEach`, which is either called by your script directly, or by the `selector.untilEach` or using the rollout manager stuff ... I'm guessing the example you provided is just the beginning part, and you have one of those calls below.

For example, with 

```
openshift.withCluster() {
    openshift.withProject() {
       openshift.selector("pod", [ app: "asdf" ]) 
    }
}
```

I get no errors.

With

```
openshift.withCluster() {
    openshift.withProject() {
       def pods = openshift.selector("pod", [ app: "asdf" ])
       pods.withEach {
         echo "GGM"
       }
    }
}
```

I get an error just like yours:  `ERROR: projects were not set for selector([name=pod],[labels={app=asdf}],[namelist=null],[projectlist=null])`

Anyway, I'll go ahead and implement the fix based on my repro.  But do let me know if what I mentioned above does not line up with what you have and we need to look at other paths.

Thanks

ramato-procon:
@gabemontero Sorry for the late answer. I didn't saw the notification.
Your fix was already released with 1.0.9, right?
Unfortunately I still get the same error.

And yes I was calling withEach.
My code including withEach:

```
openshift.withCluster(crp) {
	openshift.withProject(ocstage) {
		timeout(10) {
                        echo 'Select application pods'
			def pods = openshift.selector("pod", [ deploymentconfig: "${app}" ])
			def pod
			if (!pods.exists()) {
					error 'No pod found'
			} else {
			// Get running pod
			echo 'Selecting pod'
			pods.withEach {
				if (it.object().status.phase == 'Running') {
						pod = it.name();
						return true;
			  } else {
				  return false;
				}
			}
	  ....
   
```
The error appears after the echo 'Selecting pod', so it is definitely the withEach.

gabemontero:
[...]

PR #132, opened 2018-03-21
properly handle emtpy selectors in withEach

gabemontero:
Fixes https://github.com/openshift/jenkins-client-plugin/issues/131

### DIFF
--- a/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
+++ b/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
@@ -1519,8 +1519,10 @@ class OpenShiftDSL implements Serializable {
         public <V> V withEach(Closure<V> body) {
             defaultProjectsForObjectList();
             List<String> names = names();
+            if (names == null || names.size() == 0)
+                return;
             List<String> projects = projects();
-            if (projects == null) {
+            if (projects == null || projects.size() != names.size()) {
                 throw new AbortException("projects were not set for " + toString());
             }
             for (int i=0; i < names.size(); i++) {
--- a/examples/jenkins-image-sample.groovy
+++ b/examples/jenkins-image-sample.groovy
@@ -120,7 +120,14 @@ try {
                     builds = nb.narrow("bc").related( "builds" )
 
                 }
-
+                
+                //make sure we handle empty selectors correctly
+                // uncomment out when v1.0.9 of the plugin is in the centos image
+                /*def nopods = openshift.selector("pod", [ app: "asdf" ])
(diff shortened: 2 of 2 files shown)


gabemontero:
bump @coreydaley @bparees - any bandwidth / interest?

bparees:
no :(

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -1519,8 +1519,10 @@ class OpenShiftDSL implements Serializable {
         public <V> V withEach(Closure<V> body) {
             defaultProjectsForObjectList();
             List<String> names = names();
+            if (names == null || names.size() == 0)
+                return;
             List<String> projects = projects();
-            if (projects == null) {
+            if (projects == null || projects.size != names.size()) {
```
coreydaley: should this be projects.size() ?

gabemontero: ahh ... good catch ... not so safe after all :-)

gabemontero: and another instance of stupid groovy ;-)

coreydaley: Is there any kind of groovy linter that we could add to the testing for these repos?

gabemontero:
On Thu, Mar 22, 2018 at 10:51 AM, Corey Daley <notifications@github.com>
wrote:

> *@coreydaley* commented on this pull request.
>
> In src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
> <https://github.com/openshift/jenkins-client-plugin/pull/132#discussion_r176449000>
> :
>
> >              List<String> projects = projects();
> -            if (projects == null) {
> +            if (projects == null || projects.size != names.size()) {
>
> Is there any kind of groovy linter that we could add to the testing for
> these repos?
>
Yeah there appears to be some, both for groovy in general and jenkins
pipelines.
A simple "groovy linter" search turns up a bunch of stuff

I just have not had the time to try to wedge them into or repo mechanics.
Certainly
feel free to hack away if you think you got the time.

Also, several IDEs are suppose to have linters for groovy.  Whether it it
switching
to them, or finally upgrading my eclipse in the hopes I get better groovy
support,
are on my list too.

I also want to move the vast majority of our logic out of
OpenShiftDSL.groovy into
a java file, and just use the groovy file as a bridge.  Even have a card
open for that one.

Issue #130, opened 2018-03-19
version 1.0.8 untilEach: watch closure threw an exception: "java.lang.IndexOutOfBoundsException: Index: 1, Size: 1; Index: 1, Size: 1".


TomSchmitz:
with version 1.0.8 the untilEach function throws an IndexOutOfBoundsException.

we use the untilEach function to wait for pods to be running after a rollout of a deploymentConfig.
during the iteration the jenkins console output shows the following error:
watch closure threw an exception: "java.lang.IndexOutOfBoundsException: Index: 1, Size: 1; Index: 1, Size: 1".
the untilEach iteration will not get the true-condition to terminate the watch on the pods to be running.

here is a code example, which runs into the error described above:
```
def dcRelatedPods = deploymentConfig.related('pods')
def podObj
dcRelatedPods.untilEach(1) {
  podObj = it.object()
  return podObj.status.phase == 'Running'
} 
```

With the version 1.0.2-1.0.7 the code runs perfectly.
Is it a new behaviour or is it a bug?

Please provide help.

Thank you 
Tom

gabemontero:
hey @TomSchmitz - sound like a bug has been introduced

I have not been able to reproduce locally, constructing test cases like your sample.

Could you possibly provide the entire stack trace of the IndexOutOfBoundsException you received?

TomSchmitz:
hello @gabemontero,

in the attached file you can see the stacktrace. classes with package starting with de.generali.jenkins are our classes which wrapps the functionality of the openshift jenkins client plugin.
[indexOutOfBoundsExceptionStatcktrace.txt](https://github.com/openshift/jenkins-client-plugin/files/1832554/indexOutOfBoundsExceptionStatcktrace.txt)

If you need more information, let me know.

Thank you.
Best regards Tom

gabemontero:
Thanks @TomSchmitz - the stack trace was very helpful, as was revealing the fact you all have extended our groovy shim (first I have heard of that ... very clever ... but not something I would say we would agree to support in a first class manner).

As it turns out, v1.0.8 introduced changes we needed to support api objects in a list spread across multiple namespaces/projects ... a usage pattern that was becoming more and more prevalent.

The multi-project work very well may have had impacts on your wrappers of the OpenShiftDSL.groovy and broken assumptions you have made  ... for example, I see you've wrappered withProject ... by name alone that could be something.

[...]

PR #138, opened 2018-04-06
redo multi project fix to punt objectList<->projectList correlation, …

gabemontero:
…employ more passive project assignment

Fixes https://github.com/openshift/jenkins-client-plugin/issues/130

this fix addresses issues introduced by the earlier mulit-project fix for single project envs,
but still maintains the ability to have multi-project creates.

on the bright side, this version simplifies to a decent extent the original multi-project fix.

NOTE: I have obtained agreement from 2 upstream users to validate the fix.  I'm going the attempt to hold merging until we have validation from them, assuming they respond relatively soon.  Also, we do now have a multi-project test case in  the PR extended tests, so we should be good there.

In both upstream users cases, they are hitting timing issues between when selectors with DCs are created, and when they attempt to iterate over pods associated with them with `openshift.withEach`.  While they can hit the issue fairly reliably with their Jenkins files, it has been much less frequent with me (I have seen it like once after many runs).

I also found an issue with the `openshift.union` method ... I plan on creating a new test case for that.

### DIFF
--- a/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
+++ b/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
@@ -537,7 +537,6 @@ class OpenShiftDSL implements Serializable {
      */
     public OpenShiftResourceSelector selector(Object kind = null, Object qualifier=null) {
         OpenShiftResourceSelector selector = new OpenShiftResourceSelector("selector", kind, qualifier);
-        selector.defaultProjectsForObjectList();
         return selector;
     }
 
@@ -583,22 +582,15 @@ class OpenShiftDSL implements Serializable {
         }
                 
         Result r = new Result(verb);
-        ArrayList<String> projectNames = new ArrayList<String>();
+        HashMap<String, String> projectNames = null;
         if (consistent) {
             if (namespace == null || namespace.trim().length() == 0)
                 namespace = currentContext.getProject();
             r = innerObjectDefAction(verb, obj, userArgs, namespace, r);
-            // maintain 1 to 1 mapping with obj list and project list,
-            // even if same project 
-            if (objList != null && objList.size() > 0) {
-                for (int i=0; i < objList.size(); i++) {
-                    projectNames.add(namespace);
-                }
-            } else {
-                projectNames.add(namespace);
-            }
         } else {
-            for (Object obj2 : objList) {
+            projectNames = new HashMap<String, String>();
+            for (int i=0; i < objList.size(); i++) {
+                Object obj2 = objList.get(i);
                 if (obj2 instanceof Map) {
                     HashMap obj2Map = new HashMap((Map)obj2);
                     HashMap obj2MapMetadata = obj2Map.get("metadata");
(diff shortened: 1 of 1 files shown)


gabemontero:
Install origin flake 

/retest

gabemontero:
Several confirmations from upstream users have arrived confirming that the manually built plugins from this branch has solved the issue for them.

/hold cancel

@openshift/sig-developer-experience bump ... any takers/folks with bandwidth for a review?

I'll wait until sometime after lunch to see if anyone can.

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -583,21 +582,13 @@ class OpenShiftDSL implements Serializable {
         }
                 
         Result r = new Result(verb);
-        ArrayList<String> projectNames = new ArrayList<String>();
+        HashMap<String, String> projectNames = null;// = new HashMap<String, String>();
```
bparees: delete the comment?

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -606,14 +597,21 @@ class OpenShiftDSL implements Serializable {
                         String ns = (String)obj2MapMetadata.get("namespace");
                         if (ns == null || ns.trim().length() == 0)
                             ns = currentContext.getProject();
+                        String kind = obj2Map.get("kind");
+                        String name = obj2MapMetadata.get("name");
                         r = innerObjectDefAction(verb, obj2, userArgs, ns, r);
-                        projectNames.add(ns);
+                        projectNames.put(kind.toLowerCase().trim()+"/"+name, ns);
```
bparees: isn't this going to have collisions if someone is creating a "build/foo" in two namespaces?

gabemontero:
stopped ci job

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
gabemontero: yep should clean that up

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
gabemontero: yep you are right

alas, more groovy pain to come as I figure out how to assess :-(

gabemontero: ok I think the change to address is minor .... will test out in the AM

gabemontero:
@bparees comments addressed

gabemontero:
separate commit

gabemontero:
install origin failure .... more instances of https://github.com/openshift/origin/issues/19293

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -1487,10 +1490,15 @@ class OpenShiftDSL implements Serializable {
...
-                    String project = projectList.get(name);
+                    // if projectList != null then names() return objectList vs. live query
+                    // add index in case same kind/name in diff projects
+                    String project = projectList.get(name+i);
                     if (project != null) {
                         projList = new HashMap<String, String>();
-                        projList.put(name, project);
+                        // we hardcode the suffix to 0 since we are only sending 
+                        // 1 item from the original nameelist down ... hence the 
+                        // suffix is no the index for the first entry in the array
```
bparees: s/nameelist/namelist/
s/no/not? now?/

gabemontero: fixed thanks

Issue #137, opened 2018-04-05
have .delete(.process("yml")) work


diamark:
running cli oc allows for creation or deletion of a processed yaml file:

```
oc process -f yml | oc create -f -
oc process -f yml | oc delete -f -
```

```
openshift.create(openshift.process(readFile(yml), someparams))  works
openshift.delete(openshift.process(readFile(yml), someparams))  fails
```

I tried variations of deletion like .process().delete() and played with selectors but so far failed to have a one line equivalent to the .create(.process()) for deletion.

The delete expects selectors, I think. As pipe links in openshift.raw() have problems with setting namespace on both sides of the pipe, I am somewhat stuck with writing the output of process to file and then pass that file to a raw delete command like this `openshift.raw('delete', '-f', "${fname}")`

gabemontero:
Hey @diamark 

Yeah, with the current implementation, this works for `create` cause because since it is returning a selector, it goes through our template post processing path.

Since `delete` only returns a result, that template post processing is not currently invoked.

And you are correct around the limitations of `raw` here.

Can you elaborate on exactly how it fails?  Is it simply that the namespace/project is not properly set when the actual `oc` invocation is made?

diamark:
The call is something like this `openshift.raw(process, -f, tpl.yml, -p, FOO=bar, |, oc, delete, -f, -)` and it fails apparently because the server and token info gets appended to the actual call and thus only apply to the second oc invocation. I worked around this by forcing the process to local and that worked somewhat. But seeing the client plugin going GA and maturing pretty fast (thanks to your and the other commiters much appreciated hard work), I started to port all raw() calls to their dsl counterparts.

So this is just a humble feature/enhancement request of low prio. It just struck me as odd that something I got used to with the oc binary is kind of complicated when using dsl. Feel free to close it as WONTFIX, I am still grateful for everything the plugin offers :-)

diamark:
Oh, that one :-D . Sure, I will fetch a sample once my corporate VPN comes back online. Probably tomorrow

diamark:
Hmm, it used to be more talkative in previous versions. Here is the output of `openshift.delete(openshift.process(readFile("${largs.tpldir}/${tpl}.yml"), this.setTplParams(largs) ))`. The setup is basically the same as with https://github.com/openshift/jenkins-client-plugin/issues/133 , I just changed the .create to .delete.

[...]

PR #139, opened 2018-04-06
fix openshift.delete so that it can handle a selector groovy map and …

gabemontero:
…convert to json/yaml

Fixes https://github.com/openshift/jenkins-client-plugin/issues/137

### DIFF
--- a/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
+++ b/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
@@ -808,12 +808,11 @@ class OpenShiftDSL implements Serializable {
         return r;
     }
 
-    public Result delete(Object... oargs) {
-        String[] args = toStringArray(oargs);
-        Result r = new Result("delete");
-        r.actions.add((OcAction.OcActionResult)script._OcAction(buildCommonArgs("delete", null, args)));
-        r.failIf("delete returned an error");
-        return r;
+    public Result delete(Object obj,Object... args) {
+        OpenShiftResourceSelector deleteSelector = objectDefAction("delete", obj, args);
+        //NOTE: more groovy suckage ... I could not pass a OpenShiftResourceSelector, even
+        // though it extends Result, into the Result(Result) contructor
+        return new Result(deleteSelector.highLevelOperation, deleteSelector.actions);
     }
 
     public Result set(Object... oargs) {
@@ -881,10 +880,15 @@ class OpenShiftDSL implements Serializable {
     public static class Result implements Serializable {
 
         public final ArrayList<OcAction.OcActionResult> actions = new ArrayList<OcAction.OcActionResult>();
-        private final String highLevelOperation;
+        public final String highLevelOperation;
 
         public Result(String highLevelOperation) {
             this.highLevelOperation = highLevelOperation;
+        }
+        
+        public Result(String highLevelOperation, ArrayList<OcAction.OcActionResult> actions) {
+            this.highLevelOperation = highLevelOperation;
+            this.actions = actions;
         }
         
         public Result(Result src) {
--- a/examples/jenkins-image-sample.groovy
+++ b/examples/jenkins-image-sample.groovy
@@ -73,7 +73,13 @@ try {
                 def objs = openshift.create( bc )
                 objs.describe()
                 openshift.delete("bc", "test")
-                
+                // switch to delete below when v1.0.10 is available in the image
+                //openshift.delete(bc)
+                openshift.create("configmap", "foo")
+                openshift.create("configmap", "bar")
+                openshift.delete("configmap/foo", "configmap/bar")
+                openshift.create("configmap", "foo")
+                openshift.delete("configmap/foo")
     
                 def template
                 if (!templateExists) {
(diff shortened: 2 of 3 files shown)


src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -808,12 +808,11 @@ class OpenShiftDSL implements Serializable {
         return r;
     }
 
-    public Result delete(Object... oargs) {
-        String[] args = toStringArray(oargs);
-        Result r = new Result("delete");
-        r.actions.add((OcAction.OcActionResult)script._OcAction(buildCommonArgs("delete", null, args)));
-        r.failIf("delete returned an error");
-        return r;
+    public Result delete(Object obj,Object... args) {
```
bparees: i'm confused by the implications of this signature change on existing callers of delete()?

i trust you but i wouldn't mind an explanation of what has actually happened here that we can introduce the new arg safely.  It looks like previously the first arg was implicitly always the "resource type" and now that is explicit?

gabemontero: Certainly / no worries ... and I wasn't sure until I verified with some test cases, but as I lay this out, I realize an additional scenario should be tried ... perhaps I did break something...

First, why I did it?  It made it consistent with the create, and it made it easier to isolate the first arg, which `objectDefAction` needed.... and dealing with `Object...` in groovy has proven a pain

Next, why it works:
a) `objDefAction` handles the different cases ... i.e. is the first art a resource type? is the first arg a list of api objs (which are in either groovy/map or json format) ?
b) maybe java is too, groovy seems to be very permissive wrt Object[] and Object....  being combined, omitted 

Now, all that said, I do think I need to go back and try more things like `openshift.delete("svc/mongodb", "dc/mongodb") ... I thought I tried that and it worked (i.e. it was existing), but I don't see it now, and I didn't add it here.

I'll do that (may wait until Monday) and report back.

gabemontero: Reporting back with some findings .... the plus side is I get to highlight the shortcomings of groovy again :-)

So, this works in Java:

```
    public void delete(Object... args) {
    }

    public void delete(Object obj,Object... args) {
    }
``` 

But not in groovy ... keeping the original delete signature, and adding a new signature, such as:

```
    public Result delete(Object... args) {
        logToTaskListener("GGM original signature ");
        String[] sarr = toStringArray(args);
        ArrayList<String> al = Arrays.asList(sarr);
        logToTaskListener("GGM with params " + al);
        Object firstObj = al.remove(0);
        return delete(firstObj, al.toArray());
    }

    public Result delete(Object obj,Object... args) {
        logToTaskListener("GGM new signature " + obj)
        OpenShiftResourceSelector deleteSelector = objectDefAction("delete", obj, args);
        //NOTE: more groovy suckage ... I could not pass a OpenShiftResourceSelector, even
        // though it extends Result, into the Result(Result) contructor
        return new Result(deleteSelector.highLevelOperation, deleteSelector.actions);
    }

```

Results in this complaint:

```
java.lang.IllegalStateException: cannot compare public com.openshift.jenkins.plugins.OpenShiftDSL$Result com.openshift.jenkins.plugins.OpenShiftDSL.delete(java.lang.Object[]) to public com.openshift.jenkins.plugins.OpenShiftDSL$Result com.openshift.jenkins.plugins.OpenShiftDSL.delete(java.lang.Object,java.lang.Object[])
	at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.GroovyCallSiteSelector.isMoreSpecific(GroovyCallSiteSelector.java:291)
	at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.GroovyCallSiteSelector.findMatchingMethod(GroovyCallSiteSelector.java:203)
	at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.GroovyCallSiteSelector.method(GroovyCallSiteSelector.java:150)
	at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxInterceptor.onMethodCall(SandboxInterceptor.java:87)
	at org.kohsuke.groovy.sandbox.impl.Checker$1.call(Checker.java:155)
	at org.kohsuke.groovy.sandbox.impl.Checker.checkedCall(Checker.java:159)
	at com.cloudbees.groovy.cps.sandbox.SandboxInvoker.methodCall(SandboxInvoker.java:17)
	at WorkflowScript.run(WorkflowScript:5)
```

So this tells me changing the signature like I have done above does not break things.  Existing invocations will reach the innards of our new method like before.

Lastly, this test, which I'll add, works:

```
        openshift.create("configmap", "foo")
        openshift.create("configmap", "bar")
        openshift.delete("configmap/foo", "configmap/bar")
```

Temporary debug shows that' `configmap/foo` is in the first param, and `configmap/bar` is the first element in the args array.

I'll upload the updated test case file in a bit to add that snippet.  I also need to update the global var help file to highlight the new capabilities for `openshift.delete`

gabemontero:
ok I've pushed the help.jelly and test case updates ... I could leave some of the tests in,  as they passed with the older version of the plugin.

/hold cancel

gabemontero:
just had another plugin PR test pass install origin ... going to start trying the various tricks in the mild hope they help

/retest

PR #150, opened 2018-04-23
fix multi namespace on new -o=name output

gabemontero:

more workarounds in plugin for new `-o=name` output in 3.10

will be just merging after any review comments vs. pr test, then validating change in https://github.com/openshift/origin/pull/19465

### DIFF
--- a/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
+++ b/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
@@ -435,7 +435,20 @@ class OpenShiftDSL implements Serializable {
         for (int i = 0; i < names.length; i++) {
             String name = names[i].trim();
             if (! name.isEmpty()) {
-                results.add(name);
+               // to simplify life, to account for new -o=name output for openshift unique api objs
+               // with group suffixes like deploymentconfig.apps.openshift.io, and chop off the "domain"
+               //TODO need to consider this path, along with the actual value of the "metadata.kind" field
+               // (it currently is still "DeploymentConfig" in 3.10) when we start seeing resources like 
+               // "deploymentconfig.extensions.k8s.io"
+               String[] kind_name = name.split("/");
+               String kind = kind_name[0];
+               String[] kind_parts = kind.split("\\.");
+               if (kind_parts.length > 1) {
+                  String nm = kind_name[1];
+                  results.add(kind_parts[0] + "/" + nm);
+               } else {
+                  results.add(name);
+               }
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -435,7 +435,17 @@ class OpenShiftDSL implements Serializable {
         for (int i = 0; i < names.length; i++) {
             String name = names[i].trim();
             if (! name.isEmpty()) {
-                results.add(name);
+               // to simplify life, to acount for new -o=name output for openshift unique api objs
```
bparees: account

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -435,7 +435,17 @@ class OpenShiftDSL implements Serializable {
...
+               // with group suffixes like deploymentconfig.apps.openshift.io, and chop off the "domain"
```
bparees: so you're going to turn "deploymentconfig.apps.openshift.io/mydc" into "deploymentconfig/mydc"

what happens when there's also a "deploymentconfig.extensions.k8s.io"?

gabemontero: we'll have to see how the actual "kind" field is represented when we have those two types of deploymentconfigs and adjust accordingly

going in the reverse direction breaks backwards compatibility and is a much more invasive change

I'm willing to add a //TODO noting this possibility

gabemontero:
preliminary response to @bparees comments pushed

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
bparees: i think the Kind fields will both say "deploymentconfig" and it's the apiversion field that will provide the disambiguation.

It's probably (hopefully) unlikely enough that we can ignore it for now, but i think end-users/administrators adding their own resource types is going to be more command and someone will probably get cute and create "deploymentconfig.mycorp.com" because they don't like the openshift one, someday.  So as long as you think there will be something we can do at that point, leaving it as a todo is ok with me.

gabemontero: ok good thanks ... yeah, added the //TODO, and want to wait and see on this one, should be able to address when needed (the answer probably will be adding the version field to the key used for projectList, but I'd rather wait until we have something real to test against then anticipate what would be needed).

Issue #149, opened 2018-04-23
Patching DC example?


djohnston89:
Hey guys,

Just wondering if I can have an example of editing an item inside a DC. I'd like to add a "name" field to a port.

```spec:
    containers:
    - image: xxxxxx
      imagePullPolicy: Always
      name: play-originv3-test
      ports:
      - containerPort: 9000
        protocol: TCP
      - containerPort: 8778
        name: jolokia  (!!! THIS IS WHAT I WANT TO ADD !! )
        protocol: TCP
      resources: {}

**Comment by djohnston89** [comment]
I got this to print out the ports object:
```def dc = openshift.selector("dc", pom.artifactId)
	def dcObject = dc.object()
	echo "${dcObject.spec.template.spec.containers[0].ports}"
```
However, the openshift.apply(object) keeps giving me different errors. Some of them from various attempts:
1. Cannot restore slice from string
2. List element types are not identical

gabemontero:
Duh ... there was in fact an example:  https://github.com/openshift/jenkins-client-plugin#need-to-update-an-object-without-replacing-it

Trying again to construct a repro in the interim

djohnston89:
[patch_error_log.txt](https://github.com/openshift/jenkins-client-plugin/files/1940121/patch_error_log.txt)

gabemontero:
ah ... ok, so you didn't originally create the obj with oc apply / create --save-config

then looks like you tried to do that in subsequent attempts, but got other errors 

So it isn't exactly the same as https://github.com/openshift/jenkins-client-plugin/issues/128

Could you send me the yaml you use to create the DC, and I'll try to work off of that

djohnston89:
```def dc = openshift.selector("dc", pom.artifactId)
def dcObject = dc.object()
dcObject.spec.template.spec.containers[0].ports = '[[containerPort:8778, protocol:TCP, name:jolokia]]'
openshift.apply(dcObject)
```
[patch_error_log2.txt](https://github.com/openshift/jenkins-client-plugin/files/1940152/patch_error_log2.txt)

djohnston89:
I create the DC with new-app:

```stage('Deploy Dev') {
steps{
script{
openshift.withCluster() {
openshift.withProject(env.DEV_PROJECT) {
if(!openshift.selector('dc', pom.artifactId).exists()){
def app = openshift.newApp(pom.artifactId + ":latest")
echo "Create route"
app.narrow("svc").expose();
[...]

PR #151, opened 2018-04-24
add some more guidance around using openshift.apply

gabemontero:
Fixes https://github.com/openshift/jenkins-client-plugin/issues/149

### DIFF
--- a/README.md
+++ b/README.md
@@ -444,6 +444,91 @@ openshift.withCluster( 'mycluster' ) {
 }
 ```
 
+However, be forewarned, controllers or other users can update objects while you are trying to 
+update objects, and you can get collision conflicts.  Consider:
+
+1. Pruning any object `status` or other optional fields you are not concerned with
+2. If pruning `java.util.Map` object returned from `object()` on selectors, you'll
+need to make sure the `remove` method is in the pipeline groovy method whitelist.
+3. When pruning, you have to make sure required fields have at least a minimal or empty
+setting, even if you are not overriding them
+
+As examples, first, here is a version where we get the selector for a Deployment Config
+we created from and `openshift.newApp` call, and then pruned various fields to avoid
+conflicts, and then updated an expected port's name:
+
+```groovy
+openshift.withCluster() {
+    openshift.withProject() {
+        def app = openshift.newApp('registry.access.redhat.com/jboss-fuse-6/fis-java-openshift')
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -444,6 +444,91 @@ openshift.withCluster( 'mycluster' ) {
...
+
+As examples, first, here is a version where we get the selector for a Deployment Config
+we created from and `openshift.newApp` call, and then pruned various fields to avoid
+conflicts, and then updated an expected port's name:
+
+```groovy
+openshift.withCluster() {
+    openshift.withProject() {
+        def app = openshift.newApp('registry.access.redhat.com/jboss-fuse-6/fis-java-openshift')
+        def dc = app.narrow('dc')
```
bparees: another option might be to re-get the dc via "oc export" which strips most of this stuff for you.

bparees:
lgtm unless you want to play w/ my export suggestion.

gabemontero:
On Tue, Apr 24, 2018 at 3:55 PM, Ben Parees <notifications@github.com>
wrote:

> *@bparees* commented on this pull request.
>
> In README.md
> <https://github.com/openshift/jenkins-client-plugin/pull/151#discussion_r183859645>
> :
>
> > +
> +1. Pruning any object `status` or other optional fields you are not concerned with
> +2. If pruning `java.util.Map` object returned from `object()` on selectors, you'll
> +need to make sure the `remove` method is in the pipeline groovy method whitelist.
> +3. When pruning, you have to make sure required fields have at least a minimal or empty
> +setting, even if you are not overriding them
> +
> +As examples, first, here is a version where we get the selector for a Deployment Config
> +we created from and `openshift.newApp` call, and then pruned various fields to avoid
> +conflicts, and then updated an expected port's name:
> +
> +```groovy
> +openshift.withCluster() {
> +    openshift.withProject() {
> +        def app = openshift.newApp('registry.access.redhat.com/jboss-fuse-6/fis-java-openshift')
> +        def dc = app.narrow('dc')
>
> another option might be to re-get the dc via "oc export" which strips most
> of this stuff for you.
>
Yeah we currently don't expose `oc export` through this DSL.  If you want
to do that and start crafting
examples for its use I'd rather it be
through a separate issue / PR / card as appropriate.

bparees:
fair enough, let's go w/ this for now.

Issue #159, opened 2018-06-04
ArrayIndexOutOfBoundsException


GolanTrevize10:
Hi, we upgraded the plugin to the latest 1.0.12 version and we are getting the error below. I downgraded to 1.0.7 and it works again. We are using OC client version v3.6.0+c4dd4cf

java.lang.ArrayIndexOutOfBoundsException: 1
	at org.codehaus.groovy.runtime.dgmimpl.arrays.ObjectArrayGetAtMetaMethod.invoke(ObjectArrayGetAtMetaMethod.java:41)
	at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:325)
	at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:1213)
	at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:1022)
	at org.codehaus.groovy.runtime.callsite.PojoMetaClassSite.call(PojoMetaClassSite.java:47)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:48)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:113)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:125)
	at com.cloudbees.groovy.cps.sandbox.DefaultInvoker.getArray(DefaultInvoker.java:57)
	at com.cloudbees.groovy.cps.impl.ArrayAccessBlock.rawGet(ArrayAccessBlock.java:21)
	at com.openshift.jenkins.plugins.OpenShiftDSL.splitNames(jar:file:/var/lib/jenkins/plugins/openshift-client/WEB-INF/lib/openshift-client.jar!/com/openshift/jenkins/plugins/OpenShiftDSL.groovy:447)
	at com.openshift.jenkins.plugins.OpenShiftDSL$OpenShiftResourceSelector.startBuild(jar:file:/var/lib/jenkins/plugins/openshift-client/WEB-INF/lib/openshift-client.jar!/com/openshift/jenkins/plugins/OpenShiftDSL.groovy:1444)
	at buildInOpenshift.call(/data/jenkins-jobs/Virtual Assistant DHL Express/jobs/dhl-express-pipeline/builds/43/libs/Openshift pipeline library/vars/buildInOpenshift.groovy:6)

gabemontero:
[...]

PR #160, opened 2018-06-04
need len check on both arrays referenced in splitNames when accountin…

gabemontero:
…g for new API behavior with -o=name

Fixes https://github.com/openshift/jenkins-client-plugin/issues/159

the PR test will still fail for the known aos-cd-job change we are making ... I'm running PR testing locally and will manually merge if passes.


gabemontero:
v 1.0.13 of the client plugin has been cut

Issue #180, opened 2018-09-19
version 1.0.17: first call of OpenShiftDSL function, e.g. raw or process, fails with java.io.IOException: Cannot run program "oc"


TomSchmitz:
Hello,

after updating jenkins-client-plugin version from 1.0.16 to 1.0.17 any OpenShiftDSL function will fail with a java.io.IOException: Cannot run program "oc" (in directory ...): error=2, No such file or directory.

Calling withCluster() and withProject() works just fine, but any other OpenShiftDSL function, e.g. raw or process,  will fail.

You can find the full stacktrace in the attached text file.

If you need any further info, let me know.
Thank you for your help.
Tom
[stacktraceIOException.txt](https://github.com/openshift/jenkins-client-plugin/files/2397334/stacktraceIOException.txt)

TomSchmitz:
hello @gomesp,

our jenkins instance is running on OS RHEL 6. We have several oc CLIs installed and configured via the jenkins-client-plugin, see attached file. We don't set the oc CLI in the PATH explicity.
![jenkins_oc_client](https://user-images.githubusercontent.com/17268360/45802568-81f2a500-bcb7-11e8-8b84-808dd24a9373.png)

We just switch to version 1.0.16 again and it is running successfully without configuring anything special for the oc CLIs.

fshehadeh:
@gomesp We saw a similar issue when upgrading from 1.0.13 to 1.0.17. We configured Jenkins with multiple version of OC and we use code like the following to update the PATH environment variable:
`		def ocDir = tool ocVersion
		echo "Using OC version: ${ocVersion}, at: ${ocDir}"

		this.script.withEnv(["PATH+OC=${ocDir}"]) {
			// verify the actual project name
			def actualProjectName = openshift.project()
			echo "Using Openshift project: ${actualProjectName}"

			// do something with it
		}
`
I verified using 'echo' that the PATH environment was updated inside the withEnv block, but I don't think it is propagating to the actual call for the "oc" command. Rolling back to 1.0.13 fixed the issue. I think this has to do with the recent change to how the "oc" command is invoked (used to rely on BourneShellScript, which might have been aware of the PATH environment)

TomSchmitz:
@gomesp , in order to confirm @fshehadeh's post, we are using the similar code to define oc tool and PATH:
`
def ocDir = steps.tool ocTool
steps.withEnv(["PATH+OC=${ocDir}"]) {
  //do something with OpenShiftDSL
[...]

Issue #177, opened 2018-09-06
Escaping parameters with spaces


gomesp:
Hi team

I am not sure if this is a bug or a call for help. 

In one of my projects I am trying to create a deployment config from a template file with parameters. To do that I run (roughly):
```
String[] templateParams = ["-p","AD_GROUP=${adGroup}"]                                                     
def template = "${TEMPLATE_PATH}/dc-webservices.yml",
openshift.process(template, templateParams)
```
The template YAML has a default value with spaces and that works:

```
  - name: AD_GROUP
    required: true
    value: Application Production Logs
```

However if I override the parameter in the Jenkinsfile (in the first snippet) with a value with spaces, I get the following error:

```
ERROR: process returned an error;
{reference={}, err=error: template "Users" could not be found, verb=process, cmd=oc --server=https://xxxxx:8443 --namespace=xxxx --token=XXXXX process -f /tmp/process1812368489278526389.markup -p AD_GROUP="All Users" -o=json , out=, status=1}
```

I have tried setting ```${value}``` to:

```
adGroup = '"All Users"' # (single quote, then double quote)
adGroup = "'All Users'" # (double quote, then single quote)
adGroup = '\"All Users\"' # (escaping double quotes)
```
Have I encountered a bug, or am I not escaping the variable correctly? I appreciate any help that could be given here.
Thanks

gomesp:
BTW, running the same command from the shell works.

coreydaley:
@gomesp Can you try setting openshift.verbose in your Jenkinsfile and then run it with the default value for AD_GROUP and look at how it gets quoted?

TomSchmitz:
Ich bin bis 17.09.2018 abwesend.
I am out of the office until 17.09.2018.

Ich werde Ihre Nachricht nach meiner Rückkehr beantworten.

In dringenden Fällen wenden Sie sich bitte an Frank Becker (AC-6606)


Hinweis: Dies ist eine automatische Antwort auf Ihre Nachricht  "Re:
[openshift/jenkins-client-plugin] Escaping parameters with spaces (#177)"
gesendet am 07.09.2018 12:59:05.

Diese ist die einzige Benachrichtigung, die Sie empfangen werden, während
diese Person abwesend ist.

gomesp:
@coreydaley Thanks for your response. The error below was from Jenkins running with Openshift in verbose mode.

For ```adGroup = '"All Users"' # (single quote, then double quote):```
```
[...]

PR #182, opened 2018-09-25
Fix env propagation

gabemontero:
Fixes https://github.com/openshift/jenkins-client-plugin/issues/180
Fixes https://github.com/openshift/jenkins-client-plugin/issues/177

@openshift/sig-developer-experience @vfreex ptal

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/util/ClientCommandBuilder.java
+++ b/src/main/java/com/openshift/jenkins/plugins/util/ClientCommandBuilder.java
@@ -5,8 +5,11 @@ import com.openshift.jenkins.plugins.OpenShift;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.logging.Logger;
 
 public class ClientCommandBuilder implements Serializable {
+    
+    private static Logger LOGGER = Logger.getLogger(ClientCommandBuilder.class.getName()); 
 
     public final String server;
     public final String project;
@@ -39,13 +42,58 @@ public class ClientCommandBuilder implements Serializable {
         this.logLevel = logLevel;
     }
 
-    private static List<String> toStringArray(List l) {
+    private List<String> toStringArray(List l) {
         ArrayList<String> n = new ArrayList<String>();
         if (l == null) {
             return n;
         }
         for (Object o : l) {
-            n.add(o.toString());
+            String s = o.toString();
+            if (wrapperInQuotes() && s.contains(" ")) {
+                // the Jenkins QuotedStringTokenizer is not quite 
+                // copacetic with template parameters defined in 
+                // groovy variables (specifically strings where you have spaces but don't have 
+                // to encapsulate with " or ').
+                // 
+                // for example:
+                /*
+                def asdf = 'all users'
+                def param = ["-p", "POSTGRESQL_USER=${asdf}"]
+                def sel = openshift.selector("templates/postgresql-ephemeral")
+                def temp = sel.object()
+                def objs = openshift.process(temp, param)                 
+                 */
+                // we do *NOT* want ot force them to say 
+                // def param = ["-p", "POSTGRESQL_USER='${asdf}'"]
+                
+                // that said, for now at least, we'll try to use the Jenkins
+                // QuotedStringTokenizer for the other " and ' scenarios, especially around exec/rsh
+                // where the entire command might be listed as 'ps ax' vs. 'ps','ax'
+                // but still adjust
+                // the input with the logic here so that it 
+                // properly handles the groovy string vars with spaces scenario 
+                // correctly.
+                
+                // So, if the arg that comes in either 
+                // a) has neither ' or "", or
+                // b) has ", but doesn not start with it, and does not start with '
+                // then wrapper with '
+                if ((!s.contains("'") && !s.contains("\"")) ||
+                   (!s.startsWith("\"") && !s.startsWith("'") && s.contains("\""))) {
+                    s = "'" + s + "'";
+                // if the arg that comes in either 
+                // c) has ', but does not start with it, and does not start with "
+                // then wrapper with "
+                } else if (!s.startsWith("\"") && !s.startsWith("'") && s.contains(("'"))) {
+                    s = "\"" + s + "\"";
+                }
+                
+                // also, I have  confirmed throught testing that the jenkins groovy
+                // parser does not allow settings like '...'...'; it has to be '..."...'
+                // or "....'..."
+                
+            }
+            n.add(s);
         }
         return n;
     }
@@ -140,6 +188,17 @@ public class ClientCommandBuilder implements Serializable {
             sb.append(" ");
         }
         return sb.toString();
+    }
+    
+    public String[] asStringArray(boolean redacted) {
+        List<String> list = buildCommand(redacted);
+        return list.toArray(new String[0]);
+    }
+    
+    public boolean wrapperInQuotes() {
+        if (verb.trim().equals("process"))
+            return true;
+        return false;
     }
 
 }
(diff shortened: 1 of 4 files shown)


src/main/java/com/openshift/jenkins/plugins/util/ClientCommandBuilder.java
```diff
@@ -45,7 +48,38 @@ public ClientCommandBuilder(String server, String project, boolean skipTLSVerify
...
+                // a) has neither ' or "", or
+                // b) has ", but doesn not start with it, and does not start with '
+                // then wrapper with '
+                if ((!s.contains("'") && !s.contains("\"")) ||
+                   (!s.startsWith("\"") && !s.startsWith("'") && s.contains("\"")))
+                    s = "'" + s + "'";
+                // if the arg that comes in either 
+                // c) has ', but does not start with it, and does not start with "
+                // then wrapper with "
+                if (!s.startsWith("\"") && !s.startsWith("'") && s.contains(("'")))
```
bparees: you are doing this check against the string you just (potentially) added quotes to.  intentional?

gabemontero: ahh ... thx, will address

gabemontero:
update for first comment pushed @bparees

bparees:
lgtm.

gabemontero:
thx @bparees 

and yep, let us let @vfreex chime in before actually merging, 

but based on my testing, his examples like

`openshift.selector('po', '|| ping google.com -c 5 > out.txt ; echo {} | tee --').object()`

should be ok.  But certainly if he wants to check out the branch and give it a go himself, let's give him some time to do that, though we do have a fair number of upstream users impacted, so let's not wait super long.

Also, once we have a new version of the plugin available for the PR tester, I do have a simple reproducer or two we can add to the test pipeline.

bparees:
> Also, once we have a new version of the plugin available for the PR tester, I do have a simple reproducer or two we can add to the test pipeline.

great, tests was gong to be my next question :)

gabemontero:
ahhh ... I forgot to validate the rsh/exec path (and @vfreex mentioned that earlier today)

OK, I'll circle back to that on Wednesday.  Might need to fall back on verb based special casing I was hoping to avoid, but we'll see ...

src/main/java/com/openshift/jenkins/plugins/util/ClientCommandRunner.java
```diff
@@ -96,7 +102,44 @@ public void checkRoles(RoleChecker checker) throws SecurityException {
...
+                if (path == null || path.length() == 0) {
+                    LOGGER.warning("PATH not properly set prior to invocation of 'oc'");
+                } else {
+                    String[] dirs = path.split(File.pathSeparator);
+                    boolean override = false;
+                    for (String dir : dirs) {
+                        File f = new File(dir);
+                        File[] fileList = f.listFiles();
+                        if (fileList != null)
+                            for (File file : fileList) {
```
vfreex: Is there any particular reason to iterate over all files in the inspected directory? Can we use some condition like `if (new File(path, prog.toString()).canExecute())`?

src/main/java/com/openshift/jenkins/plugins/util/ClientCommandBuilder.java
```diff
@@ -45,7 +48,39 @@ public ClientCommandBuilder(String server, String project, boolean skipTLSVerify
             return n;
         }
         for (Object o : l) {
-            n.add(o.toString());
+            String s = o.toString();
+            if (s.contains(" ")) {
+                // the Jenkins QuotedStringTokenizer is not quite 
+                // copacetic with template parameters defined in 
+                // groovy variables (specifically strings where you have spaces but don't have 
+                // to encapsulate with " or ').
```
vfreex: I am not sure if I fully understand this. Can you list some examples?

src/main/java/com/openshift/jenkins/plugins/util/ClientCommandRunner.java
```diff
@@ -96,7 +102,44 @@ public void checkRoles(RoleChecker checker) throws SecurityException {
...
+                                if (file.isFile() && file.getName().equals("oc")) {
```
vfreex: Does OpenShift Client plugin support Windows slaves? If yes, I think `oc.exe`, `oc.bat`,etc  (see `PATHEXT` environment variables)  should also be checked.

src/main/java/com/openshift/jenkins/plugins/util/ClientCommandRunner.java
```diff
@@ -96,7 +102,44 @@ public void checkRoles(RoleChecker checker) throws SecurityException {
...
+                // 2) invoke a shell that then launches the actual 'oc' command, where we update the PATH prior
+                // Our use of the jenkins durable task and bourne/windows scripts did 2), but because of 
+                // https://bugzilla.redhat.com/show_bug.cgi?id=1625518 we analyze the PATH env var and do 1)
+                String path = envVars.get("PATH");
+                // default if running the openshift jenkins images
+                String dirToUse = "/bin/";
+                if (path == null || path.length() == 0) {
+                    LOGGER.warning("PATH not properly set prior to invocation of 'oc'");
+                } else {
+                    String[] dirs = path.split(File.pathSeparator);
```
vfreex: I remember that there is a short circuit: Linux and Windows searches `PATH` only when the specified command doesn't contain `/` (or `\` on Windows).

src/main/java/com/openshift/jenkins/plugins/util/ClientCommandRunner.java
gabemontero: Good catch @vfreex - I've never used it, but there is a windows client.  It is `oc.exe` btw.

I'll update.

src/main/java/com/openshift/jenkins/plugins/util/ClientCommandRunner.java
gabemontero: Interesting ... in our case it should not matter though.  The command passed to this point will always be `oc` or `oc.exe` ... there will be no fully qualified path to oc.

src/main/java/com/openshift/jenkins/plugins/util/ClientCommandRunner.java
gabemontero: Yeah I like that ... I'll give it a go and assuming it works will change to that form.

src/main/java/com/openshift/jenkins/plugins/util/ClientCommandBuilder.java
gabemontero: I'll see about adding some more specific examples in the comment.

And yes, limited use of `QuotedStringTokenizer`, or more machinations here, are the paths I'm considering to deal with the `rsh` / `exec` scenarios vs. the `ASDF=${asdf}` scenario during template parameter passing.

We technically do have a spec, albeit example driven, which just lacked precision wrt these groovy nuances with our API.  I don't see us investing in a formalized spec for this beyond the current help.jelly, example based approach.  Perhaps we start adding warning messages when we detect `ASDF=${asdf}` in the parameter passing that the value should be quoted, and that we'll start strictly enforcing that at a later date, but we'll have to live with whatever short term hack I land on for the time being.  We'll see ...

gabemontero:
changes pushed ... also they crossed the 2 commits enough that I just collapsed it all in one commit

gabemontero:
Saw the oc run failure that just happened in my 3.9 testing, but thought it was due to an old oc I was running .... but it is in fact related to my fix ... updating coming soon I believe.

gabemontero:
ok fully passing for me locally now, including added test case along with existing test cases

gabemontero:
and ok tests passing now

I'll give @bparees and @vfreex a bit to see the latest change before merging 

And I have the updated test case ready once the new version is in the centos image

vfreex:
I am OK with this as long as it fix the current problem we met. Thanks @gabemontero .

gabemontero:
v1.0.18 of the plugin has been initiated with the jenkins update center

PR #188, opened 2018-10-08
add verify service functionality provided by old plugin

gabemontero:
Decided to add as part of moving samples off of old plugin to this plugin

@coreydaley @adambkaplan @bparees ptal

### DIFF
--- a/examples/jenkins-image-sample.groovy
+++ b/examples/jenkins-image-sample.groovy
@@ -199,6 +199,8 @@ void actualTest() {*/
                     dcs.describe()
                     echo "DeploymentConfig history"
                     dcs.rollout().history()
+
+                    //openshift.verifyService('mongodb')
     
                     def rubySelector = openshift.selector("bc", "ruby")
                     def builds
--- a/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
+++ b/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
@@ -15,6 +15,10 @@ import java.io.IOException
 import java.util.logging.Level
 import java.util.logging.Logger
 
+import java.net.InetSocketAddress;
+import java.net.Socket;
+
+
 import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution;
 import org.jenkinsci.plugins.workflow.cps.CpsThread;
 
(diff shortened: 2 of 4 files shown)


src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -905,6 +909,38 @@ class OpenShiftDSL implements Serializable {
         return newObjectsAction("startBuild", "start-build", args);
     }
 
+    public boolean verifyService(String svcName) {
+        OpenShiftResourceSelector svcSel = this.selector("svc", svcName);
+        String ip = (String)svcSel.object().get("spec").get("clusterIP");
+        String port = (String)svcSel.object().get("spec").get("ports").get(0).get("port");
+        int i = 0;
+        while (i < 5) {
+            i++;
```
adambkaplan: use a for loop instead?

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -905,6 +909,38 @@ class OpenShiftDSL implements Serializable {
...
+            InetSocketAddress address = new InetSocketAddress(ip, Integer.parseInt(port));
+            Socket socket = null;
+            try {
+                socket = new Socket();
+                socket.connect(address, 2500);
```
adambkaplan: current maximum execution time is 12.5 seconds - is this too long?

examples/jenkins-image-sample.groovy
```diff
@@ -199,6 +199,8 @@ void actualTest() {*/
                     dcs.describe()
                     echo "DeploymentConfig history"
                     dcs.rollout().history()
+
+                    //openshift.verifyService('mongodb')
```
adambkaplan: Any reason to keep this commented out?

gabemontero: we cannot actually add tests until *after* the pr merges and a new version of the plugin is created, otherwise it breaks our overnight tests

minor drawback with sharing the PR tests and overnight tests that we just live with

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
gabemontero: it was what we did with the old plugin

and users can override with the `timeout(...)` pipeline step enclosing a call to our new method here

adambkaplan: catching the interrupt exception makes more sense with that timeout

PR #189, opened 2018-10-09
enable OcAction --follow support

gabemontero:
see https://github.com/openshift/jenkins-client-plugin/issues/55#issuecomment-428172800

another somewhat smaller hiccup from the refactor off of durable task plugin

### DIFF
--- a/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
+++ b/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
@@ -434,6 +434,22 @@ class OpenShiftDSL implements Serializable {
         String verb = toSingleString(overb);
         String[] userArgsArray = toStringArray(ouserArgsArray);
         String[] overrideArgs = toStringArray(ooverrideArgs);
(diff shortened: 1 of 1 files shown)


src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -434,6 +434,21 @@ class OpenShiftDSL implements Serializable {
...
+                streamStdOutToConsolePrefix = verb;
+                break;
+            }
+        }
+        if (streamStdOutToConsolePrefix == null)
+            for (String arg : overrideArgs) {
+                String a = arg.trim();
+                if (a.equals("-f") || a.equals("--follow")) {
+                    streamStdOutToConsolePrefix = verb;
+                }
```
adambkaplan: missing `break`?

gabemontero: yep

gabemontero: thx!

Issue #191, opened 2018-10-15
Jenkins keeps creating threads in the executor thread pool while executing the pipelines and does NOT close or release the threads after the pipeline execution has completed.


mattsjk:
**ENVIRONMENT DETAILS:**
- Virtual Machine running RedHat Enterprise 7.0
- Java JVM – OpenJDK 1.8.0_151-b12
- Jenkins LTS version 2.138.1 with Openshift Client plugin 1.0.19

We run mostly multibranch pipelines and have 250 executors with 1 slave to pick up the long running tasks.

The build stages will checkout the code from a bitbucket repository and build using maven and then create the deploy image to a cloud container using OpenShift v3.9.30 - all with DSL based scripts within the pipeline.

**ISSUE/PROBLEM:**
Jenkins keeps creating threads in the executor thread pool while executing the pipelines and does NOT close or release the threads after the pipeline execution has completed.

The threads created by the executor thread pool go into a WAITING state and are never released.

After a certain limit to the number of threads, we have to restart Jenkins to reset the thread count. A thread count graph from the monitoring page (JavaMelody) is shown below:

![image](https://user-images.githubusercontent.com/12805032/46978667-0e17b100-d12c-11e8-81dc-609801502317.png)

Checking the thread details using the monitoring (JavaMelody) plugin – there are 2 threads each for a thread pool like: "pool-xxxx-thread-1" & "pool-xxxx-thread-2" and this is the details for one of the thread:
**pool-100-thread-1**
_sun.misc.Unsafe.park(Native Method)
java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2039)
java.util.concurrent.LinkedBlockingQueue.take(LinkedBlockingQueue.java:442)
java.util.concurrent.ThreadPoolExecutor.getTask(ThreadPoolExecutor.java:1074)
java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1134)
java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
java.lang.Thread.run(Thread.java:748)_

**TEMPORARY SOLUTION:**
[...]

PR #192, opened 2018-10-16
do not create executor pool on each command execution

gabemontero:
Fixes https://github.com/openshift/jenkins-client-plugin/issues/191


gabemontero:
@openshift/sig-developer-experience

gabemontero:
v1.0.21 of the client plugin has been initiated

PR #198, opened 2018-10-31
version lock the template being used

bparees:


bparees:
@gabemontero ptal, part of why our origin 3.10 e2e tests are failing is because origin 3.10 doesn't have the same imagestreams as 3.11, and the upstream templates our tests are instantiating assume all the latest imagestreams are present.

This jenkinsfile is utilized by one of the tests, so I need to fix it here so I can then point the test to this branch.

bparees:
(this is ugly and in the future we should probably decouple our tests from the upstream templates for cases where the upstream templates are not the thing we are testing)

bparees:
i dunno what testing would run against this since it's not against the master branch, i'm just going to merge it.

gabemontero:
was going to mentioned that (test/master branch) but then got sidetracked ... glad you sorted it out quick enough

PR #207, opened 2018-12-17
fix multicontainer agent pipelines (no ClientCommandRunner, but no Ba…

gabemontero:
…sh durable task)

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1657208


gabemontero:
v1.0.23 of the plugin has these commits

Issue #209, opened 2018-12-21
Retry command fails and gets stuck inside a loop


cloudowski:
We have the following piece of code in our pipeline:

```
timeout(time: 3, unit: 'MINUTES') {
                    podSelector.untilEach {
                        echo "pod: ${it.name()}"
                        return it.object().status.containerStatuses[0].ready
                    }
                }
```

It checks whether all containers are ready. In most cases, it's a single container, but sometimes our job fails and it seems like it's not retrying check of the container status - it seems like it's not entering loop.
We get the following output:

```
11:39:36 Checking watch output again in 300ms
11:39:36 Checking watch output again in 360ms
11:39:36 Checking watch output again in 432ms
11:39:37 Checking watch output again in 518ms
11:39:37 Checking watch output again in 621ms
11:39:38 Checking watch output again in 745ms
11:39:39 Checking watch output again in 894ms
11:39:40 Checking watch output again in 1072ms
11:39:41 Checking watch output again in 1286ms
11:39:42 Checking watch output again in 1543ms
11:39:43 Checking watch output again in 1851ms
11:39:45 Checking watch output again in 2221ms
11:39:47 Checking watch output again in 2665ms
11:39:50 Checking watch output again in 3198ms
11:42:18 Cancelling nested steps due to timeout
```

It should retry in 3.2s, but it didn't and it finally timed out. It sometimes work, but recently we noticed increased number of such errors.
We use version 1.0.23.

szpak:
We took a closer look at the issue and the problem appears on reading data from `/tmp/watchstderr...txt` file. When to the initial content of that file (e.g. `pod/foobar-28-nxsjq`) the same  line is added on deployment completion the execution seems to hang on:
```
byte[] newOutput = Files.readAllBytes(path);
```
It's terminated by our timeout a few minutes later. The deployment itself finishes successfully.

Looking at the recent changes it seems to be a regression introduced in https://github.com/openshift/jenkins-client-plugin/commit/11ba99797511157f9c8dfea18bfec6f34ed6edfc ([bug 1657208](https://bugzilla.redhat.com/show_bug.cgi?id=1657208)) (1.0.23). The problem was not observable with 1.0.22. 

@gabemontero Maybe you have any thought on that?

Jenkins 2.121.3. OpenShift 3.11.16.

gabemontero:
Thanks for the additional detail @szpak 

[...]

PR #214, opened 2019-01-17
get back to essential pre 1.0.17 file processing for watch/action

gabemontero:
Fixes https://github.com/openshift/jenkins-client-plugin/issues/209

@openshift/sig-developer-experience fyi ... as I mentioned in scrum today, in addition to some upstream users seeing these intermittent issues, I believe I've seen an instance or two of this (blocking on the `Files.readAllBytes` call) in some recent `aws-builds` runs on CI

These change more closely approximate what we use to do in this space prior to our abandoning the durable task / bourne shell stuff that proved untenable for a set of template param / env var based `oc` scenarios

### DIFF
--- a/examples/multicontainer-pipieline.groovy
+++ b/examples/multicontainer-pipieline.groovy
@@ -1,0 +1,41 @@
+pipeline {
+    agent {
+      kubernetes {
+        cloud 'openshift'
+        label 'mypod'
+        defaultContainer 'jnlp'
+        yaml """
+apiVersion: v1
+kind: Pod
+metadata:
+  labels:
+    some-label: some-label-value
+spec:
+  serviceAccount: jenkins
+  containers:
+  - name: maven
+    image: docker.io/openshift/jenkins-agent-maven-35-centos7:v4.0
+    command:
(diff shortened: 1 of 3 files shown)


gabemontero:
@bparees fyi ... after CI still passing over the last week, in the last day, the ansible setup for the plugin test has been failing with:

```
TASK [Run variable sanity checks] **********************************************
task path: /usr/share/ansible/openshift-ansible/playbooks/init/sanity_checks.yml:14
fatal: [localhost]: FAILED! => {
    "generated_timestamp": "2019-01-25 17:14:14.795801", 
    "msg": "last_checked_host: localhost, last_checked_var: openshift_master_identity_providers;Found removed variables: openshift_node_labels is replaced by openshift_node_groups[<item>].labels; "
}

```

I have not asked those guys (at least yet), but initial suspicion is that openshift-ansible has changed enough with 4.0 that these old test paths may be untenable.

If so, we'll be back to manually testing and green button merges for some of these upcoming fixes to the plugins until we can sort out the new ci strategy and plugins.

Issue #217, opened 2019-01-29
verifyService does not work with headless Services


jtudelag:
verifyService does not work with headless Services.

Headless Services do not have clusterIPs associated, so when the method verifyService try to parse the IP, it fails.

https://github.com/openshift/jenkins-client-plugin/blob/release-3.10/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy#L946

gabemontero:
interesting use case @jtudelag 

Longer term, some new support to add.  Short term, we can highlight this req/assumption in the help/readme where needed.

waveywaves:
I would like to work on this issue @gabemontero.
Anything I need to keep in mind before I start this ?

gabemontero:
working out how to test the headless service from within groovy / java will be the tricky part here @waveywaves 

the queries you have to make into openshift for headless require more sophistication and parsing

ultimately, if you parse the api objects correctly, you get an ip /port which you can test via socket connection like the current verifyService functionality does

and of course, you need to be able to handle either headless and non-headless services from the same method.

the details around headless service are at https://docs.openshift.com/container-platform/3.11/architecture/core_concepts/pods_and_services.html#headless-services

and before you ask ;-) .... I've never set up a headless service before ... so you get to ask question on the openshift mailing lists if you have trouble

TomSchmitz:
Ich bin bis 03.06.2019 abwesend.
I am out of the office until 03.06.2019.

Ich werde Ihre Nachricht nach meiner Rückkehr beantworten.

In dringenden Fällen wenden Sie sich bitte an Frank Becker (AC-6606)


Hinweis: Dies ist eine automatische Antwort auf Ihre Nachricht  "Re:
[openshift/jenkins-client-plugin] verifyService does not work with headless
Services (#217)" gesendet am 01.06.2019 08:12:18.

Diese ist die einzige Benachrichtigung, die Sie empfangen werden, während
diese Person abwesend ist.

PR #258, opened 2019-04-16
Verify Headless Services

waveywaves:
Added ability to verify headless service using the verifyService() function

Fixes https://github.com/openshift/jenkins-client-plugin/issues/217

cc @gabemontero

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@
 - [Installing and developing](#installing-and-developing)
 - [Compatibility with Declarative Pipeline](#compatibility-with-declarative-pipeline)
 - [Compatibility with parallel step](#compatibility-with-parallel-step)
+- [Compatibility with load step](#compatibility-with-load-step)
 - [Examples](#examples)
   - [Hello, World](#hello-world)
   - [Centralizing Cluster Configuration](#centralizing-cluster-configuration)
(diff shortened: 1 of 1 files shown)


src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -945,28 +947,47 @@ class OpenShiftDSL implements Serializable {
...
-            } catch (IOException e) {
-                logToTaskListener("IOException " + e.getMessage() + " connecting to service " + svcName);
-                LOGGER.log(Level.FINE, "verifyService", e);
-                try {
-                    Thread.sleep(2500);
-                } catch (InterruptedException e1) {
+            if (ip == "None"){ 
+                InetAddress inetAddress = null;
+                try{
+                    inetAddress = InetAddress.getByName(svcName);
```
gabemontero: @waveywaves I've got a question in to our networking team on whether the dns lookup should work even if the jenkins pod is in a separate namespace than the headless service

gabemontero: I think it should be want to make sure.

In the interim, if you could construct such a test case and confirm/deny that fact, do that and respond here.

waveywaves: I will check if it works with services in other namespaces as well once and update it here along with the README.

PR #261, opened 2019-04-27
Updated verifyService to support headless services with selectors

waveywaves:
Closed PR https://github.com/openshift/jenkins-client-plugin/pull/258 by mistake.
Fixes https://github.com/openshift/jenkins-client-plugin/issues/217

##### Update:
I used the `dig` command as follows from a pod in one namespace to query a headless service in another which worked. 
```sh
dig nodejs-ex-headless.jenkins-headless-test.svc.cluster.local A +short                                                                           
10.130.1.173
10.131.0.74
10.131.0.75
```
We get the IPs of the pods above. 

If I use `openshift.verifyService` to test the same service, we get the following error below
```
ERROR: Unable to retrieve object markup with get; action failed: {reference={}, err=Error from server (NotFound): services "nodejs-ex-headless.jenkins-headless-test.svc.cluster.local" not found
, verb=get, cmd=oc --server=https://172.30.0.1:443 --certificate-authority=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt --namespace=vibhav-jenkins --token=XXXXX get service/nodejs-ex-headless.jenkins-headless-test.svc.cluster.local -o=json , out=, status=1}
```

cc @gabemontero @sthaha

### DIFF
--- a/README.md
+++ b/README.md
@@ -659,7 +659,7 @@ openshift.withCluster() {
 ```
 
 If you are looking for the equivalent of `openshiftVerifyService` from [the OpenShift Jenkins Plugin](https://github.com/openshift/jenkins-plugin), we added a similar operation.
-
+This works with Services with ClusterIP as well as [headless Services (with selectors)](https://kubernetes.io/docs/concepts/services-networking/service/#with-selectors) in the namespace specified by the openshift.withProject().
 ```groovy
 openshift.withCluster() {
     openshift.withProject() {
--- a/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
+++ b/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
@@ -12,12 +12,16 @@ import hudson.FilePath
 import hudson.Util
 
 import java.io.IOException
+import java.lang.management.BufferPoolMXBean
 import java.util.logging.Level
 import java.util.logging.Logger
 
 import java.net.InetSocketAddress;
 import java.net.Socket;
-
+import java.net.InetAddress;
+import java.net.UnknownHostException
+import java.util.regex.Matcher;
+import java.util.regex.Pattern
 
 import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution;
 import org.jenkinsci.plugins.workflow.cps.CpsThread;
@@ -942,31 +946,51 @@ class OpenShiftDSL implements Serializable {
     }
 
     public boolean verifyService(String svcName) {
+
         OpenShiftResourceSelector svcSel = this.selector("svc", svcName);
         String ip = (String)svcSel.object().get("spec").get("clusterIP");
         String port = (String)svcSel.object().get("spec").get("ports").get(0).get("port");
+        InetAddress inetAddress = null;
+        String hostAddress = null;
         int i = 0;
         while (i < 5) {
             i++;
-            InetSocketAddress address = new InetSocketAddress(ip, Integer.parseInt(port));
-            Socket socket = null;
-            try {
-                socket = new Socket();
-                socket.connect(address, 2500);
-                logToTaskListener("Connected to service " + svcName + " via IP " + ip + " and port " + port);
-                return true;
-            } catch (IOException e) {
-                logToTaskListener("IOException " + e.getMessage() + " connecting to service " + svcName);
-                LOGGER.log(Level.FINE, "verifyService", e);
-                try {
-                    Thread.sleep(2500);
-                } catch (InterruptedException e1) {
+            // ClusterIp is not present in headless services with selectors.
+            if (ip.trim() == "None"){
+                try{
+                    inetAddress = InetAddress.getByName(svcName+"."+svcSel.project());
+                    logToTaskListener("Connected to headless service " + svcName + " via port " + port + " with pod IP " +  inetAddress.getHostAddress());
+                    return true;
                 }
-            } finally {
+                catch (UnknownHostException e){
+                    logToTaskListener("UnknownHostException " + e.getMessage() + " connecting to service " + svcName);
+                    LOGGER.log(Level.FINE, "verifyService", e);
+                    try {
+                        Thread.sleep(2500);
+                    } catch (InterruptedException e1) {
+                    }
+                }
+            } else {
+                InetSocketAddress address = new InetSocketAddress(ip, Integer.parseInt(port));
+                Socket socket = null;
                 try {
-                    if (socket != null)
-                        socket.close();
+                    socket = new Socket();
+                    socket.connect(address, 2500);
+                    logToTaskListener("Connected to service " + svcName + " via IP " + ip + " and port " + port);
+                    return true;
                 } catch (IOException e) {
+                    logToTaskListener("IOException " + e.getMessage() + " connecting to service " + svcName);
+                    LOGGER.log(Level.FINE, "verifyService", e);
+                    try {
+                        Thread.sleep(2500);
+                    } catch (InterruptedException e1) {
+                    }
+                } finally {
+                    try {
+                        if (socket != null)
+                            socket.close();
+                    } catch (IOException e) {
+                    }
                 }
             }
         }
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -658,7 +658,8 @@ openshift.withCluster() {
 }
 ```
 
-If you are looking for the equivalent of `openshiftVerifyService` from [the OpenShift Jenkins Plugin](https://github.com/openshift/jenkins-plugin), we added a similar operation.
+If you are looking for the equivalent of `openshiftVerifyService` from [the OpenShift Jenkins Plugin](https://github.com/openshift/jenkins-plugin), we added a similar operation. This works with Services with ClusterIP as well as [headless Services (with selectors)](https://kubernetes.io/docs/concepts/services-networking/service/#with-selectors).
+The svcName has to be provided in the following format : serviceName.namespace.svc.cluster.local
```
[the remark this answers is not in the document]
waveywaves: The expected behavior from verifyService is same as what `dig` does in the pod and sucessfully verifies the service, but that is not what we see. Tried this out while running `dig` from inside a `giantswarm/tiny-tools` container. In this case the `tiny-tools` container and the `jenkins` container were in the same namespace and the similar behaviour was expected from `verifyService('serviceName.namespace.svc.cluster.local')` in the `jenkins` container and `dig` in the `tiny-tools` container. The service in another namespace and the namespace which was being verified was created manually. Could that be the reason it didn't work ?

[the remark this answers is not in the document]
gabemontero: Hmmm ... you should be able to use `oc rsh..` to get into your jenkins pod.  Do that and see what happens when you run `dig` from within the jenkins pod.

Based on the results, that could take us in a couple of different directions.

Your question around service setup could very well have bearing, but I don't know exactly how yet.

Re-reading https://docs.okd.io/latest/architecture/core_concepts/pods_and_services.html#headless-services my guess is there is some delta we are missing with respect to the DNS setup that is described in that selection ... with selectors / without selectors.

I'm also wondering about differences between the JDK and what `dig` etc. do.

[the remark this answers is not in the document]
waveywaves:
```
sh-4.2$ dig nodejs-ex-headless.jenkins-headless-test.svc.cluster.local A +short
10.130.1.173
10.131.0.74
10.131.0.75
```

The above command was run from the jenkins pod in the `vibhav-jenkins` namespace querying the headless service `nodejs-ex-headless` in the `jenkins-headless-test` namespace.

Seems like `dig` is able to access but JDK is not. 
Whereas JDK is able to access services in the local namespace.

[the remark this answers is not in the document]
gabemontero: yep this is what I was afraid of in my earlier speculations

it has to be some sort of env difference where the DNS server lookup is different within the JVM vs. the /etc/resolve.conf which ultimately gets mounted in the pod and which `dig` leverages

I did a search on "java dns resolver configuration" and got some hits like https://stackoverflow.com/questions/11647629/how-to-configure-hostname-resolution-to-use-a-custom-dns-server-in-java where folks did custom setup of the DNS lookup path within the JVM

I would not complain if you dropped this given the difficulties, but for educational purposes, if you are so inclined, you could 
a) `oc rsh` back into the pod and `cat /etc/resolv.conf` to get the list of DNS servers 
b) modify the java code to employ one of the examples from https://stackoverflow.com/questions/11647629/how-to-configure-hostname-resolution-to-use-a-custom-dns-server-in-java to either examine or set the DNS lookup to ensure they are consistent, and then see what happens with the lookup.

gabemontero:
so yeah address the outstanding comments @waveywaves and then narrow the scope via doc and help to single namespace

waveywaves:
@gabemontero The necessary changes have been made.
Do check this and let me know.

README.md
```diff
@@ -659,6 +659,7 @@ openshift.withCluster() {
 ```
 
 If you are looking for the equivalent of `openshiftVerifyService` from [the OpenShift Jenkins Plugin](https://github.com/openshift/jenkins-plugin), we added a similar operation.
+This works with Services with ClusterIP as well as [headless Services (with selectors)](https://kubernetes.io/docs/concepts/services-networking/service/#with-selectors) in the same namespace.
```
gabemontero: `in the same namesapce.` -> `in the namespace specified by the openshift.withProject()` closure.`

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -942,31 +946,52 @@ class OpenShiftDSL implements Serializable {
...
-                socket.connect(address, 2500);
-                logToTaskListener("Connected to service " + svcName + " via IP " + ip + " and port " + port);
-                return true;
-            } catch (IOException e) {
-                logToTaskListener("IOException " + e.getMessage() + " connecting to service " + svcName);
-                LOGGER.log(Level.FINE, "verifyService", e);
-                try {
-                    Thread.sleep(2500);
-                } catch (InterruptedException e1) {
+            if (ip == "None"){
```
gabemontero: `oc` is known to change  ... we might always get "None"

how comfortable are you constructing a regular expression check or some other validation that determines if it is a valid ipv4 / ipv6 address ?

src/main/resources/com/openshift/jenkins/plugins/pipeline/OpenShiftGlobalVariable/help.jelly
```diff
@@ -206,6 +206,10 @@
             <code id="openshift_verifyService">openshift.verifyService(svc_name:String):boolean</code>
         </dt>
         <dd>
+            <p>
+                Allows you to check if your Openshift service is reachable or not. This works with headless Services (with Selectors)
+                as well as normal Services with ClusterIP which are in the same namespace.
```
gabemontero: same `openshift.withProject()` adjustment to this message

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
akram: @gabemontero did you mean:
oc is known to change ... we might **NOT**  always get "None"

I am not sure about this, because setting None as a cluster IP is the explicit way to define a headless service as stated in [1].

If it is not the case, then @waveywaves, the check needs to be simplify. Could be useful to use Apache commons-validator which has an InetAddressValidator.isInetAddress method.


If we don't have this dependency, which seems to be the case, at least, we need to extract constants of the regex.

[1] https://kubernetes.io/docs/concepts/services-networking/service/#headless-services

gabemontero:
On Thu, Jul 4, 2019 at 3:27 AM Akram Ben Aissi <notifications@github.com>
wrote:

> *@akram* commented on this pull request.
>
> In src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
> <https://github.com/openshift/jenkins-client-plugin/pull/261#discussion_r300312085>
> :
>
> >          int i = 0;
>          while (i < 5) {
>              i++;
> -            InetSocketAddress address = new InetSocketAddress(ip, Integer.parseInt(port));
> -            Socket socket = null;
> -            try {
> -                socket = new Socket();
> -                socket.connect(address, 2500);
> -                logToTaskListener("Connected to service " + svcName + " via IP " + ip + " and port " + port);
> -                return true;
> -            } catch (IOException e) {
> -                logToTaskListener("IOException " + e.getMessage() + " connecting to service " + svcName);
> -                LOGGER.log(Level.FINE, "verifyService", e);
> -                try {
> -                    Thread.sleep(2500);
> -                } catch (InterruptedException e1) {
> +            if (ip == "None"){
>
> @gabemontero <https://github.com/gabemontero> did you mean:
> oc is known to change ... we might *NOT* always get "None"
>

Yep that was what I meant but given the k8s link you found I retract my
comment about regex

Let’s do a trim space on the string and a ignore case string compare

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -942,31 +948,51 @@ class OpenShiftDSL implements Serializable {
...
-            } catch (IOException e) {
-                logToTaskListener("IOException " + e.getMessage() + " connecting to service " + svcName);
-                LOGGER.log(Level.FINE, "verifyService", e);
-                try {
-                    Thread.sleep(2500);
-                } catch (InterruptedException e1) {
+            // ClusterIp is not present in headless services with selectors.
+            if (ip.trim() == "None"){
+                try{
+                    inetAddress = InetAddress.getByName(svcName);
```
gabemontero: change to `InetAddress.getByName(svcName+"."+svcSel.project())`

gabemontero:
This will hit the same bluegreen nodejs issue the jenkins repo is hitting
on the next round of PR test I believe (unless the plugins are not based on
4.2 - don’t remember offhand)

gabemontero:
Must have accepted the prior test runs.

PR #264, opened 2019-05-07
remove space param test again since we have 3.11 jenkins e2e working …

gabemontero:
…again and haven't decided about bumping the version of client plugin in 3.11


gabemontero:
We are seeing the old `Unauthorized` bug ... see https://github.com/openshift/release/pull/3710 to getting the right ocp cluster base to test against.

PR #267, opened 2019-05-22
travis/prow no longer copacetic

gabemontero:


adambkaplan:
@gabemontero I'd make sure a verify job is added in openshift/release (if you haven't done so already)

gabemontero:
@adambkaplan could you post an example of what you are referring to?

adambkaplan:
@gabemontero https://github.com/openshift/release/blob/master/ci-operator/config/openshift/origin/openshift-origin-master.yaml#L241-L246 from origin

Doesn't have to be as fancy

gabemontero:
ah ... I see what you mean now .. thanks for the link @adambkaplan 

turns out though, the `mvn clean package` done at https://github.com/openshift/jenkins-client-plugin/blob/master/Dockerfile#L8 during the build seciont of our ci flow is a superset of the `mvn clean verify` ... at least that is what I see when I compare the two.

So I'm going to refrain on adding more churn in openshift/release.

PR #287, opened 2019-08-08
Pipeline CPS method mismatch: Add proper CPS handing for the jenkins-client-plugin

akram:
- Turn contextContains method into @openshift/openshift-architects 
- fix bug with dieIfNotWithin which does the opposite check

### DIFF
--- a/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
+++ b/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
@@ -96,6 +96,7 @@ class OpenShiftDSL implements Serializable {
         ContextId(String name) {
             this.@name = name;
         }
+        @NonCPS
         public String toString() {
             return name;
         }
@@ -137,6 +138,7 @@ class OpenShiftDSL implements Serializable {
             }
         }
 
+        @NonCPS
         public ContextId getContextId() {
             return this.@id;
         }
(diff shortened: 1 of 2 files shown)


akram:
/assign @gabemontero 
/assign @waveywaves 

@gabemontero I have seen this in interactive session with Vibhav, maybe you can make a review also.
We are going to submit addition e2e tests in origin and also a Jenkins Plugin update for workflow-cps-plugin to 2.73

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -267,7 +270,7 @@ class OpenShiftDSL implements Serializable {
 
     @NonCPS
     private void dieIfNotWithin(ContextId me, Context context, ContextId required) throws AbortException {
-        if (contextContains(context, required)) {
+        if (!contextContains(context, required)) {
```
gabemontero: yeah ... the fix seems right  ... a bit surprised this did not come up sooner ... prior to your change here that line had not been touched since Oct 25 2017

good catch!!

I'm half wondering if people have been getting away with a withProject only, where somehow it defaulted to project jenkins is running in, and will get bitten by your change.

Some simple test cases added to openshift/origin or even at least the examples dir here for reference wouldn't be a bad idea, just to get the warm fuzzy ... also would be good to compare the results of that simple test case before and after ... assuming it would reproduce what issue you and @waveywaves were hitting @akram

PR #302, opened 2019-09-27
Add an example for using openshift.tag

itwasonlyabug:
I got left scratching my head if this is supported or not until I tried it so I wanted to add it to the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -683,16 +683,31 @@ That step was introduced prior to the introduction of the OpenShift Pipeline Bui
 With the advent of OpenShift Pipeline Build Strategy, incorporating your pipeline into such a BuildConfig along with the use of an Image Change Trigger is the better choice for triggering pipeline jobs from changes to ImageStreams in OpenShift.
 
 ### Tagging images across namespaces
+
+Re-tagging existing images or promoting images across different namespaces (e.g. from Staging to Production) can be done easily and only requires valid credentials.
+
+```groovy
+openshift.withCluster( 'mycluster' ) {
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -680,8 +681,19 @@ No equivalent of the ImageStream SCM (SCM steps are a special extension point in
 That step was introduced prior to the introduction of the OpenShift Pipeline Build Strategy.
 
 With the advent of OpenShift Pipeline Build Strategy, incorporating your pipeline into such a BuildConfig along with the use of an Image Change Trigger is the better choice for triggering pipeline jobs from changes to ImageStreams in OpenShift.
-   
 
+### Tagging images across namespaces
```
waveywaves: @itwasonlyabug Thank you for the PR :)

Can you also add a little but if verbose description as to what you are doing in this example ? 


/assign waveywaves

waveywaves:
@akram and I are the maintainers now so we would be taking care of the bugs please refrain from assigning reviewers by yourself.

/unassign @adambkaplan

README.md
itwasonlyabug: Hi! Do you mean as in the [description under the heading here](https://github.com/openshift/jenkins-client-plugin#sticking-with-the-defaults) or with more comments?

I want to keep it consistent with the rest, but the style varies a little :)

I could also add other examples for tagging images, but I'm not sure if the use-case would be helpful.

waveywaves: Description under the header should be good.

waveywaves: Looks good.

PR #308, opened 2019-10-23
Fix 306 namespace not passed to raw command

### LINKED ISSUE #306: raw command does not accept namespace as a argument
According to https://docs.openshift.com/container-platform/3.9/admin_guide/pruning_resources.html#pruning-deployments

When I'm running this command in command line I get multiple results for deployments to be pruned:

```
oc adm prune deployments --keep-younger-than=168h --keep-complete=0 --keep-failed=0
```

When I do the same with jenkins client plugin I get 0 results. Specifying namespace doesn't  help.

```
openshift.raw( "adm prune deployments", --keep-younger-than=168h",
                "--keep-complete=0 --keep-failed=0 -n test-project")
```

It does work for pruning builds, however. But not deployments.

akram:
Fixes #306 

Allows to pass verb arguments to openshift.raw() command, especially the "-n namespace" or "--namespace=value" to be able to override the default namespace.

Adds also a few formatting.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,0 +1,21 @@
+#!/bin/sh
+
+# Deploys locally the plugin in a pod which has label name=jenkins
+# You must be correctly logged in an OpenShift or Kubernetes cluster (KUBECONFIG set or oc login)
+pod_name=$( oc get pods -l name=jenkins --output=name | cut -f2 -d/)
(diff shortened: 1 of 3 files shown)


akram:
/assgin @waveywaves 

Can you PTAL ?

Issue #326, opened 2020-04-29
Bug 1837362: Sensitive data exposed on pipeline failure


rinormaloku:
When creating a secret using this plugin, `oc apply` failures will show the output in the Jenkins Job output. This can lead to leakage of sensitive data. E.g.

Using this task:
OpenShift - Create Resource(s)
```
apiVersion: v1
kind: Secret
metadata:
  name:  secretname
data:
   secretKey: YWRtaW4K
   secret: xxxxK
type: Opaque
```

Will fail with the following output:
```
Started by user anonymous
Building in workspace /Users/rmaloku/work/jenkins-client-plugin/work/jobs/test/workspace
Executing: oc --server=https://kubernetes.docker.internal:6443 --certificate-authority= --namespace=default --token=XXXXX create -f /var/folders/r5/dh4_6p_91r90wkxmkw957g0w0000gn/T/markup4721899811229862180.tmp 
Error from server (BadRequest): error when creating "/var/folders/r5/dh4_6p_91r90wkxmkw957g0w0000gn/T/markup4721899811229862180.tmp": Secret in version "v1" cannot be handled as a Secret: v1.Secret.Data: decode base64: illegal base64 data at input byte 4, error found in #10 byte of ...|t":"xxxxK","secretKe|..., bigger context ...|{"apiVersion":"v1","data":{"secret":"xxxxK","secretKey":"YWRtaW4K"},"kind":"Secret","metadata|...
Client tool terminated with status: 1
Build step 'OpenShift - Create Resource(s)' marked build as failure
Finished: FAILURE
```

This being the important line: *{"secret":"xxxxK","secretKey":"YWRtaW4K"},"kind":"Secret","metadata|...*

This can be fixed by one of the two ways:
1. Reading the entire output and then regex-ing and redacting the sensitive data. 
2. Or have a flag that asap some sensitive keywords are noticed (e.g. "Secret") turns off logging to the command line.

I'd prefer the first approach as we would be able to provide the user with more details without leaking sensitive data.

If you see this as a beneficial addition to the plugin, please share your suggestions with me and I will work on it.

akram:
/retitle Bug 1837362: Sensitive data exposed on pipeline failure

PR #327, opened 2020-05-07
Bug 1837362: Sensitive data exposed on pipeline failure

rinormaloku:
Resolves the ticket #326 

i.e. the following output:
```
{err=Error from server (BadRequest): error when creating "/Users/rmaloku/work/jenkins-client-plugin/work/workspace/test-ocp/apply2288494789001734078.markup": Secret in version "v1" cannot be handled as a Secret: v1.Secret.Data: decode base64: illegal base64 data at input byte 4, error found in #10 byte of ...|t":"xxxxK","secretKe|..., bigger context ...|{"apiVersion":"v1","data":{"secret":"xxxxK","secretKey":"YWRtaW4K"},"kind":"Secret","metadata|...
``` 
is printed as:
```
{err=Error from server (BadRequest): Secret in version "v1" cannot be handled as a Secret: v1.Secret.Data: decode base64: illegal base64 data at input byte 4, error found in #10 byte of ...|t":"xxxxK","secretKe|..., bigger context ...|{"apiVersion":"v1","data":{ REDACTED },"kind":"Secret","metadata|...
```

The important section is: **"data":{ REDACTED }**

FYI: I didn't hide the following section `|t":"xxxxK","secretKe|` as it provides the user with information of where the mistake is. Though I am open for your suggestions if we want to hide this as well.

### DIFF
--- a/src/main/java/com/openshift/jenkins/plugins/freestyle/BaseStep.java
+++ b/src/main/java/com/openshift/jenkins/plugins/freestyle/BaseStep.java
@@ -14,17 +14,20 @@ import hudson.Launcher;
 import hudson.model.AbstractBuild;
 import hudson.model.AbstractProject;
 import hudson.model.Computer;
-import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.tasks.BuildStepDescriptor;
 import hudson.tasks.Builder;
 import hudson.util.ListBoxModel;
 
+import org.apache.commons.io.IOUtils;
+import org.apache.commons.io.LineIterator;
 import org.kohsuke.stapler.DataBoundSetter;
 import org.kohsuke.stapler.QueryParameter;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -34,6 +37,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
+
+import static com.openshift.jenkins.plugins.util.ClientCommandOutputCleaner.redactSensitiveData;
 
 public abstract class BaseStep extends Builder {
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/openshift/jenkins/plugins/freestyle/BaseStep.java
```diff
@@ -255,16 +260,18 @@ public boolean perform(ProcessBuilder pb)
                         new Thread(new Runnable() {
                             @Override
                             public void run() {
-                                byte buffer[] = new byte[1024];
-                                int count;
-                                try {
-                                    while ((count = output.read(buffer)) != -1) {
-                                        listener.getLogger().write(buffer, 0,
-                                                count);
+                                StringBuilder sb = new StringBuilder();
```
akram, line 263: StringBuilder is not synchronized, so, we may endup in log messages displayed in a different order from what they are produced.
Any objections in using StringBuffer instead ?

src/main/java/com/openshift/jenkins/plugins/pipeline/OcAction.java
```diff
@@ -63,14 +64,18 @@ public OcAction(String server, String project, boolean skipTLSVerify, String caP
         public int status;
         @Whitelisted
         public HashMap<String, String> reference = new HashMap<String, String>();
+        @Whitelisted
+        public boolean verbose = false;
```
akram, line 68: should this flag be propagated to all the commands in the GroovyDSL?
if so, we can do it in a separate PR.

rinormaloku: I forwarded it from the Step verbose property, to ensure that OcActionResult doesn't print the entire yaml/json on Jenkins. 
But I might misunderstand the question and would love suggestions on how to improve it.

src/main/java/com/openshift/jenkins/plugins/freestyle/BaseStep.java
rinormaloku: @akram Thanks for the correction, I was not aware of the difference.

akram:
/test e2e-aws-jenkins

looks like an infra flake, static pods cannot listen on port 6443

akram:
/retitle Bug 1837362: Sensitive data exposed on pipeline failure

rinormaloku:
Hi Jitendar,

To test the freestyle I just created one resource:

apiVersion: v1
kind: Secret
metadata:
  name:  secretname
data:
   secretKey: YWRtaW4K
   secret: xxxxK
type: Opaque

This should fail and hide the secret YWRtaW4K i.e. it should not be visible.

testing the pipeline was a little harder, I created an OpenShift template
with the above resource and then applied the template with
openshift.apply().

I'd provide all the scripts, but sadly my jenkins workplace is reset and I
have lost all those.

Best,
Rinor

PR #335, opened 2020-06-21
move openshift/origin using examples/jenkins-image-sample.groovy into repo

gabemontero:
OK @akram @waveywaves @adambkaplan @bparees here is the openshift CVO operator styled e2e in this repo for testing out this plugin against a cluster pointed to by your KUBECONFIG env var.

I was able the reproduced the problem seen in openshift/origin build e2e with the problematic portion of https://github.com/openshift/jenkins-client-plugin/blob/master/examples/jenkins-image-sample.groovy#L275-L299 around that second jenkins deployment which I removed a week or so ago which then got put back in ... I used this to verify the job log and pod dump on error elements of our e2e's in openshift/openshift which I ported here, but of course with the ginkgo/omega/k8s e2e framework stuff.

I also verified commenting out the second deployment stuff lets the pipeline pass ... that is what is pushed here

this resolves a long standing pain point @bparees and I discussed in past, and which was one of the avalanche of data points I dumped on @akram and @waveywaves last year

namely, testing against a PR's version of the test pipeline vs. the master branch version of the test pipeline, which has undoubtedly hit us again recently, and bugs in that pipeline got introduced

I found it was easier to move the https://github.com/openshift/jenkins-client-plugin/blob/master/examples/jenkins-image-sample.groovy into a string var in the e2e's vs. continuing to use https://github.com/openshift/jenkins-client-plugin/blob/master/examples/jenkins-image-sample.groovy

But @akram @waveywaves you can play with that as you take this over.

Ultimately,
- we are going to remove all direct use of https://github.com/openshift/jenkins-client-plugin/blob/master/examples/jenkins-image-sample.groovy from openshift/origin build e2e.... all maintained release, over the next few days
- when the dust settles, I'll help with getting an openshift/release config/job set up so running this test will be part of openshift/jenkins-client-plugin PRs ... and it will pull in the PRs version of the plugin
- I can also help with updating the makefile to build a jenkins image with your local plugin and update the jenkins imagestream in your KUBECONFIG cluster if you all like dev/test this way

Check out the branch @akram @waveywaves and give it a whirl


gabemontero:
@bparees 's PR to skip the client plugin tests  in openshift/origin master branch e2e build tests just merged

My PR https://github.com/openshift/origin/pull/25179 will re-enable the remaining client plugin tests, but without the https://github.com/openshift/jenkins-client-plugin/blob/master/examples/jenkins-image-sample.groovy test, which will now be covered by these chagnes

akram:
ack @gabemontero 
@waveywaves and @jkhelil will know both about it, and we will merge it.

test/e2e/ref.go
```diff
@@ -0,0 +1,46 @@
...
+func NewRef(t *testing.T, kubeClient *kubeset.Clientset, testNamespace string) *JenkinsRef {
+	svc, err := kubeClient.CoreV1().Services(testNamespace).Get(context.Background(), "jenkins", metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("%#v", err)
+	}
+	serviceIP := svc.Spec.ClusterIP
+	port := svc.Spec.Ports[0].Port
+
+	j := &JenkinsRef{
+		host:       serviceIP,
```
akram, line 32: I saw this also in the current e2e test, and I was wondering why and how the service IP is used in the logs. Usually, we see the jenkins Route host there; I remember that I click the link there to go watching in the logs in the jenkins blueocean console in my browser.

akram:
merging this as this as it blocking other PR's .
Let's discuss on slack about  openshift/origin#25179 and the CVO usage 

/lgtm

test/e2e/ref.go
gabemontero: We do a `curl` with the jenkins SA token to download the job log from the Jenkins instance.  When done inside a pod running in the namespace, you can use the service IP vs. the route.

Of course with the service IP, you avoid having to go over the public internet, and you avoid having to worry about DNS being properly set up *IN* the cluster to resolve that address.
The curl runs in a pod that runs in the same namespace as jenkins.

But yes, when you click the link from your browser, you laptop's DNS comes into play and things resolve.

gabemontero:
With this merged @akram I'll see this afternoon about creating the requisite config/jobs in openshift/release so future PRs for this repo have to execute the new test.

It should look at log like the config/jobs for openshift cvo operators, hopefully.

@bparees @adambkaplan FYI

akram:
awesome

akram:
/retitle Bug 1847421: move openshift/origin using examples/jenkins-image-sample.groovy into repo

PR #336, opened 2020-06-22
Revert fix on openshift.run('jenkins-second-deployment')

akram:


examples/jenkins-image-sample.groovy
```diff
@@ -289,14 +289,14 @@ void actualTest() {*/
```
gabemontero, line 289: are we sure lines 275 to 288 still work @akram ?  what are the details on what you've tested ?

gabemontero:
OK @akram tells me on slack that manual verification of this file passes

/lgtm

after this merges I'll initiate some e2e-gcp-builds runs in https://github.com/openshift/origin/pull/25171

PR #369, opened 2021-09-30
Fix oc command line when paths contains spaces

nfalco79:
Some variables that point to file can have space, for example in the name of the Jenkins job. In this case the command line produce a error in the process execution. Surrounding the path with quote solve the issue.


nfalco79:
these changes are already in our production enviroment

PR #371, opened 2021-10-07
'restart' subcommand is missing for rollout()

### LINKED ISSUE #370: 'restart' subcommand is missing for rollout()
A new rollout subcommand 'restart' appeared with OpenShift 4.x but can't be used with current openshift-client plugin.

I then tried to run it this way:
`openshift.raw("rollout status sts ${sts}")`
Unfortunately I got following error: `flags cannot be placed before plugin name`

Here is the full debug log:
```
Verbose sub-step output:
	Command> oc --server=https://10.238.0.1:443 --certificate-authority=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt --namespace=test --loglevel=8 --token=XXXXX  oc rollout restart sts/myapp
	Status> 1
	StdOut>
	StdErr> flags cannot be placed before plugin name: --server=https://10.238.0.1:443
	Reference> {}
```

Apparently, to get the new subcommand to work, only this line of code have to be added:
`public Result restart(Object...args) throws AbortException { return runSubVerb("restart", args); }`
Here in this file: https://github.com/openshift/jenkins-client-plugin/blob/master/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy#L1825

Do you confirm the fix ?

ThomVivet:
Fix #370

### DIFF
--- a/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
+++ b/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
@@ -1822,6 +1822,7 @@ class OpenShiftDSL implements Serializable {
         public Result history(Object...args) throws AbortException { return runSubVerb("history", args, true); }
         public Result latest(Object...args) throws AbortException { return runSubVerb("latest", args); }
         public Result pause(Object...args) throws AbortException { return runSubVerb("pause", args); }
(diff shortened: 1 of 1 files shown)


ThomVivet:
@akram / @sbose78, can you review this PR please ?

akram:
can you squash the commits @ThomVivet or remove the unused one?

PR #381, opened 2022-02-25
Use parent credentials class to allow Jenkins Sync credentials to be used

akram:
Use parent credentials class to allow Jenkins Sync credentials to be used side by side with the Jenkins Client plugins
Additional fix
Fixes https://github.com/openshift/jenkins-sync-plugin/issues/414

### DIFF
--- a/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
+++ b/src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
@@ -2,6 +2,7 @@ package com.openshift.jenkins.plugins
 
 import com.cloudbees.groovy.cps.NonCPS
 import com.cloudbees.plugins.credentials.CredentialsProvider
+import com.cloudbees.plugins.credentials.impl.BaseStandardCredentials
 import com.openshift.jenkins.plugins.pipeline.OcAction
 import com.openshift.jenkins.plugins.pipeline.OcContextInit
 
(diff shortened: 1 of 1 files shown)


src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
```diff
@@ -145,7 +146,7 @@ class OpenShiftDSL implements Serializable {
 
         public String getToken() {
             if (this.@credentialsId != null) {
-                OpenShiftTokenCredentials tokenSecret = CredentialsProvider.findCredentialById(credentialsId, OpenShiftTokenCredentials.class, script.$build(), Collections.emptyList());
+                BaseStandardCredentials tokenSecret = CredentialsProvider.findCredentialById(credentialsId, BaseStandardCredentials.class, script.$build(), Collections.emptyList());
```
gabemontero, line 149: please add a comment here explaining why we need the parent class here but then use `OpenShiftTokenCredentials` elsewhere

gabemontero (changes requested):

src/main/resources/com/openshift/jenkins/plugins/OpenShiftDSL.groovy
akram: comments added

gabemontero:
/label qe-approved

will open bz for the openshift/jenkins PR that pulls in the new vesrion

petr-muller:
@gabemontero @akram Tide is stuck trying to merge this because you have it [configured to use the rebase method](https://github.com/openshift/release/blob/master/core-services/prow/02_config/openshift/jenkins-client-plugin/_prowconfig.yaml#L11) but the PR cannot be rebased because of conflict (it can be merged but not rebased).

(not surfacing it well is a Tide bug https://github.com/kubernetes/test-infra/issues/25088)

PR #385, opened 2022-03-17
JNKS-267: check JENKINS-IMAGE existence, tag into jenkins imagestream, from Makefile test-e2e

gabemontero:
since the client plugin  itself does not need to use the CI step registry, a copy the work from https://github.com/openshift/release/pull/27009 

after this merges, I'll update the step-registry and see if we can use the new script defined here

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ all: test-e2e
 
 
 test-e2e:
+	./hack/tag-ci-image.sh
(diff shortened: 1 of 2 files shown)


gabemontero:
/label docs-approved

CI only change

gabemontero:
/label qe-approved

CI only change

gabemontero:
/label px-approved

CI only change

PR #387, opened 2022-04-05
RHDEVDOCS-3941 Replace plugin with plug-in

rolfedh:
Aligned team: Dev Tools
- For branches: 
- Jira: https://issues.redhat.com/browse/RHDEVDOCS-3941
- SME review: @akram


akram (changes requested):
the README.md in the vendor directory must not be changed.
vendor files must be kept untouched as they are automatically injected from go lang source dependency management.
can you please revert this one?

Apart from that, all the rest lgtm.

